PR #77, opened 2026-02-03
Implement the speech-to-text training pipeline

jodavis:
Use DVC to create a reproducible pipeline to generate speech samples and train a neural ML model for speech recognition, which can be tuned to the individual user and command space for Adaptive Remote.


jodavis:
@copilot  address the code review comments in this PR.

PR #96, opened 2026-02-23
[ADR-127] Programmable IR Commands

jodavis:
Working branch for [ADR-127](https://jodasoft.atlassian.net/browse/ADR-127): Programmable IR Commands

[ADR-127]: https://jodasoft.atlassian.net/browse/ADR-127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jodavis:
@copilot There's some flakiness in the end to end tests in this branch. There's a NullReferenceException coming from the LifecycleViewController, specifically in the anonymous callback method passed to LastOrDefault. That callback is very simple, the only thing that could be null is the item in the activities list, but nothing should be adding nulls to that list. There are locks around list changes so it shouldn't be a threading issue. Take a look at the code, run the tests, and try to get all the tests in the repo passing. Make sure the whole solution builds without warnings or errors when you are done.

PR #146, opened 2026-04-19
[ADR-177] Mapping from CompiledLayout DTO to internal view models

jodavis:
Implements client-side consumption of CompiledLayout by mapping layout DTOs (AdaptiveRemote.Contracts) into the runtime RemoteLayoutElement model tree used by the app, with updated stub orchestration and unit tests.

### DIFF
--- a/src/AdaptiveRemote.App/Configuration/HostBuilderExtensions.cs
+++ b/src/AdaptiveRemote.App/Configuration/HostBuilderExtensions.cs
@@ -29,7 +29,7 @@ internal static class HostBuilderExtensions
         => services
             .AddApplicationLifecycleServices()
             .AddCloudAssetServices()
-            .AddScoped(sp => sp.GetRequiredService<ICloudAssetStore>().Get<CompiledLayout>("layout"))
+            .AddScopedCloudAsset<CompiledLayout>("layout")
             .AddScopedLifecycleService<LifecycleCommandService>()
             .AddScoped<IRemoteDefinitionService, RemoteLayoutDefinitionService>()
             .AddScoped<IDynamicStylesheetProvider, LayoutStylesheetProvider>()
--- a/src/AdaptiveRemote.App/Configuration/HostBuilderExtensions.cs
+++ b/src/AdaptiveRemote.App/Configuration/HostBuilderExtensions.cs
@@ -1,4 +1,5 @@
-﻿using AdaptiveRemote.Services;
+﻿using AdaptiveRemote.Contracts;
+using AdaptiveRemote.Services;
 using AdaptiveRemote.Services.CloudAssets;
 using AdaptiveRemote.Services.Commands;
 using AdaptiveRemote.Services.Layout;
@@ -28,6 +29,7 @@ internal static class HostBuilderExtensions
         => services
             .AddApplicationLifecycleServices()
             .AddCloudAssetServices()
+            .AddScoped(sp => sp.GetRequiredService<ICloudAssetStore>().Get<CompiledLayout>("layout"))
             .AddScopedLifecycleService<LifecycleCommandService>()
             .AddScoped<IRemoteDefinitionService, RemoteLayoutDefinitionService>()
             .AddScoped<IDynamicStylesheetProvider, LayoutStylesheetProvider>()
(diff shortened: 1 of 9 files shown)


src/AdaptiveRemote.App/Configuration/HostBuilderExtensions.cs
```diff
@@ -28,6 +29,7 @@ internal static IServiceCollection AddRemoteServices(this IServiceCollection ser
         => services
             .AddApplicationLifecycleServices()
             .AddCloudAssetServices()
+            .AddScoped(sp => sp.GetRequiredService<ICloudAssetStore>().Get<CompiledLayout>("layout"))
```
jodavis: Add a helper extension method that takes a generic parameter and asset name, and move it to the CloudAssetHostBuilderExtensions.

src/AdaptiveRemote.App/Services/Layout/RemoteLayoutDefinitionService.cs
```diff
@@ -1,37 +1,84 @@
...
 /// </summary>
 internal class RemoteLayoutDefinitionService : IRemoteDefinitionService
 {
-    private readonly ICloudAssetStore _store;
+    private readonly Lazy<RemoteLayoutElement> _remoteRoot;
 
     public RemoteLayoutDefinitionService(ICloudAssetStore store)
     {
-        _store = store;
+        _remoteRoot = new Lazy<RemoteLayoutElement>(() => BuildRoot(store));
```
jodavis: Instead of using Lazy, just call BuildEoot right away. It's definitely going to be needed. If it fails, the whole app will fail.

src/AdaptiveRemote.App/Services/Layout/RemoteLayoutDefinitionService.cs
```diff
@@ -1,37 +1,84 @@
...
     }
 
-    public RemoteLayoutElement RemoteRoot
+    public RemoteLayoutElement RemoteRoot => _remoteRoot.Value;
+
+    private static RemoteLayoutElement BuildRoot(ICloudAssetStore store)
```
jodavis: Just import CompiledLayout instead of ICloutAssetStore. It should already be registered as a scoped service

src/AdaptiveRemote.App/Services/Layout/RemoteLayoutDefinitionService.cs
jodavis: This change was not made. Please fix.

src/AdaptiveRemote.App/Services/Layout/RemoteLayoutDefinitionService.cs
jodavis: This change was not made. Please fix.

src/AdaptiveRemote.App/Configuration/HostBuilderExtensions.cs
jodavis: This was partially fixed. The new helper was added but it is not being used here.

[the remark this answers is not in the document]
jodavis: This will be resolved when CompiledLayout is being consumed directly instead of going through ICloudAssetStore.

src/AdaptiveRemote.App/Services/Layout/RemoteLayoutDefinitionService.cs
```diff
@@ -1,37 +1,84 @@
...
+                "Ensure CloudAssetOrchestrator has completed initialization before the first scope is created.",
+                ex);
         }
+
+        List<RemoteLayoutElement> elements = layout.Elements
+            .Select(MapElement)
+            .ToList();
+        elements.Add(BuildGutter());
+
+        return new LayoutGroup("ROOT", elements);
```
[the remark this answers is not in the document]
jodavis: This is not fixed.

src/AdaptiveRemote.App/Services/Layout/RemoteLayoutDefinitionService.cs
```diff
@@ -1,37 +1,84 @@
...
+        {
+            CommandType.TiVo => new TiVoCommand(
+                cmd.Name, placement: null, label: cmd.Label,
+                cssid: cmd.CssId, glyph: cmd.Glyph, reverse: cmd.Reverse,
+                speakName: cmd.SpeakPhrase),
+
+            CommandType.IR => new IRCommand(
+                cmd.Name, placement: null, label: cmd.Label,
+                cssid: cmd.CssId, glyph: cmd.Glyph, reverse: cmd.Reverse,
+                speakName: cmd.SpeakPhrase),
```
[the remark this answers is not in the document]
jodavis: Looks like this was partially fixed. There's a new speakName and speakPhrase difference, but we're using speakName here which will still duplicate the speech.

PR #148, opened 2026-04-19
[ADR-178] ICloudAsset abstraction, JsonCloudAsset<T>, and stub file downloader

jodavis:
Introduce the asset abstraction and replace in-code CompiledLayout construction with JSON deserialization from a file, so the stub orchestrator exercises the same parse path as the real orchestrator will.


src/AdaptiveRemote.App/Services/CloudAssets/CloudSettings.cs
jodavis: We have duplication between CloudSettings and BackendSettongs. Wet need to consolidate these two.

src/AdaptiveRemote.Headless/dev/layout.json
jodavis: We should use enum names for "type" instead of integers

test/AdaptiveRemote.App.Tests/Services/CloudAssets/JsonCloudAssetTests.cs
```diff
@@ -0,0 +1,51 @@
...
+{
+    private static JsonCloudAsset<CompiledLayout> MakeSut() =>
+        new("layout", "/stream", "layout-ready", "/layouts/compiled",
+            LayoutContractsJsonContext.Default);
+
+    [TestMethod]
+    public async Task JsonCloudAsset_ParseAsync_CorrectlyDeserializesCompiledLayoutAsync()
+    {
+        // Arrange
+        CompiledLayout expected = new(
```
jodavis: Use a mock object for these unit tests instead of CompiledLayout. It will make them less fragile if CompiledLayout changes, and will validate that the JsonCloudAsset works with any type.

PR #153, opened 2026-04-20
 [ADR-179] ApplicationLifecycle recycle loop and BlazorAppScope.RecycleAsync

jodavis:
Convert ApplicationLifecycle.ExecuteAsync to a recycle loop and implement BlazorAppScope.RecycleAsync().


scripts/setup-playwright-sandbox.sh
jodavis: I would like to find a better way to support this. I'll do some research.

PR #156, opened 2026-04-22
[ADR-180] Idle detection: IIdleDetector, IdleDetector, and three ViewModel adapter services

jodavis:
Introduces token-based idle detection and wires it up to the three ViewModel sources that indicate user activity:

- IIdleDetector / IdleDetector (Services/CloudAssets/): singleton; StartNonIdle() returns an IDisposable hold; when all holds are released a cooldown timer starts; BecameIdle fires and IsIdle becomes true after CloudSettings.IdleCooldownSeconds. New StartNonIdle() during cooldown cancels the pending timer.

- ConversationIdleAdapter (Services/Layout/): scoped IScopedLifecycle; subscribes to ConversationView.IsListening and holds a non-idle token while true.

- ProgrammingModeIdleAdapter (Services/Layout/): scoped IScopedLifecycle; subscribes to LifecycleView.IsProgrammingMode and holds a non-idle token while true.

- CommandExecutionIdleAdapter (Services/Layout/): scoped IScopedLifecycle; subscribes to Command.IsActive on every command from IRemoteDefinitionService and holds one token per active command.

All three adapters re-subscribe on each scope recycle via IScopedLifecycle and release their subscriptions on CleanUpAsync.

IdleDetector registered as singleton IIdleDetector in AddCloudAssetServices. Adapters registered via AddScopedLifecycleService<T> in AddRemoteServices.

Unit tests cover: default idle state, token hold/release, cooldown timing, cooldown reset, multiple tokens, idempotent dispose, and all adapter property-change scenarios. All 373 unit tests and 14 E2E tests pass.

https://claude.ai/code/session_0132p39nh2gxTcLMPae2nNf2


src/AdaptiveRemote.App/Services/CloudAssets/IIdleDetector.cs
jodavis, line 1: Move this to the `AdaptiveRemote.Services` namespace. That's where contract interfaces consumed by other services are kept.

src/AdaptiveRemote.App/Services/CloudAssets/IIdleDetector.cs
```diff
@@ -0,0 +1,11 @@
+namespace AdaptiveRemote.Services.CloudAssets;
+
+// Tracks whether the user is idle. Non-idle state is held open by calling StartNonIdle();
+// the returned IDisposable releases the hold when disposed. When all holds are released,
+// a cooldown timer starts; BecameIdle fires and IsIdle becomes true after the cooldown.
+internal interface IIdleDetector
+{
+    bool IsIdle { get; }
+    event EventHandler? BecameIdle;
+    IDisposable StartNonIdle();
```
jodavis: Add XML comments for contract members.

src/AdaptiveRemote.App/Services/CloudAssets/IdleDetector.cs
```diff
@@ -0,0 +1,88 @@
...
+    private readonly object _lock = new();
+    private int _activeTokenCount;
+    private Timer? _cooldownTimer;
+
+    public IdleDetector(IOptions<CloudSettings> settings)
+    {
+        _cooldown = TimeSpan.FromSeconds(settings.Value.IdleCooldownSeconds);
+    }
+
+    public bool IsIdle { get; private set; } = true;
```
jodavis: We should start out with IsIdle=false, and wait for the first cooldown period before considering the application idle.

src/AdaptiveRemote.App/Services/Layout/ConversationIdleAdapter.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        _remoteDefinition = remoteDefinition;
+        _idleDetector = idleDetector;
+    }
+
+    public string Name => "Conversation idle adapter";
+
+    public Task InitializeAsync(ILifecycleActivity activity, CancellationToken cancellationToken)
+    {
+        _view = _remoteDefinition.GetElement<ConversationView>();
```
jodavis: We can do this once, in the constructor. The remote definition is expected to be consistent across a given scope.

src/AdaptiveRemote.App/Services/Layout/ConversationIdleAdapter.cs
jodavis, line 1: There's a lot of common code in the adapters. I wonder if we could create a base class (using INotifyProperty or MvvmObject as a generic "adaptable" view model) to share more functionality. It seems the only thing that needs to change is a way to determine whether the view model is "active" (non-idle) and all the rest is the same.

CommandExecutionIdleAdapter would have to be split into two components, an adapter for a single command and an IScopedLifecycle that creates single-command adapters for every command. That's better separation of responsibilities anyway.

jodavis, line 1: Move each adapter to the service namespace it relates to. (This should be in AdaptiveRemote.Services.Conversation)

src/AdaptiveRemote.App/Services/Layout/CommandExecutionIdleAdapter.cs
jodavis, line 1: Move each adapter to the service namespace it relates to. (This should be in AdaptiveRemote.Services.Commands)

src/AdaptiveRemote.App/Services/Layout/ProgrammingModeIdleAdapter.cs
jodavis, line 1: Move each adapter to the service namespace it relates to. (This should be in AdaptiveRemote.Services.ApplicationLifecycle)

test/AdaptiveRemote.App.Tests/Services/Layout/IdleAdapterTests.cs
jodavis, line 1: Once most of the functionality of the idle adaptors is moved to a common base class, we can focus on testing that by creating a TestIdleAdapter that derives from the base class. Then we don't have to have separate tests for the individual adapter classes, since "is active" detection should be extremely simple.

PR #199, opened 2026-05-23
GH-120: Cancel in-progress learning when a new button is selected

### LINKED ISSUE #120: It is possible to put two buttons into learning mode at the same time.
**Describe the bug**
After clicking on a button to start learning, it is possible to click on a different button and have both buttons learning at the same time.

**To Reproduce**
When I click on the 'Learn' button
And I click on the 'Mute' button
And I click on the 'Power' button
Then I should see the 'Power' button is active
And I should see the 'Mute' button is not active

**Expected behavior**
Clicking on a different button should cancel learning for the first button and start learning for the new button.

jodavis-claude:
## Work item

GH-120 — When a user clicks a programmable button while another button's learning cycle is already running, the running cycle should be cancelled and the new button's learning should start immediately.

## Changes

- **`BroadlinkCommandService.cs`** — Replaced `WaitAsync(0)` early-return guard with a cancel-and-wait pattern: added `_currentLearnCts` field; on each `ProgramAsync` invocation, the previous CTS is exchanged via `Interlocked.Exchange` and cancelled before waiting on the semaphore; `learnCts.Token` is linked into the body's `linkedCts`; timeout-catch condition widened to exclude cancellation from `learnCts`; `_currentLearnCts` cleared atomically in `finally` via `Interlocked.CompareExchange`.
- **`MessageLogger.cs`** — Removed EventId 1009 (`LearningAlreadyInProgress`, which described the old silently-drop behaviour); added EventId 1010 (`CancellingLearningForNewCommand`) at `LogLevel.Information`.
- **`BroadlinkCommandServiceTests.cs`** — Replaced `WhileLearningAlreadyInProgress_ReturnsImmediately` (which verified the wrong behaviour) with `WhileOtherLearningInProgress_CancelsPreviousAndStartsNew` and `DuplicateClick_CancelsPreviousAndStartsSelf`.
- **`ProgrammingMode.feature`** (WPF host) — Added scenario "Programming a command is preempted by clicking a different programmable button": clicks Mute, verifies modal, clicks Power, verifies Power modal replaces it, sends IR data, verifies Power is programmed and Mute is not.
- **`ServiceFixture.cs`** — Changed `Directory.Exists` to `Path.Exists` when walking for the repo root, fixing a test-infrastructure failure in git worktree checkouts.

## Design decisions

- **`_currentLearnCts` set before semaphore wait (not after):** This ensures a concurrent second request can always find and cancel the first request's CTS, even if the first request is still waiting for the semaphore. Three rapid clicks (A→B→C) result in A and B both cancelled and C completing — acceptable "last click wins" semantics.
- **Disposal ownership:** `learnCts` is a `using`-scoped CTS owned by the operation that created it. `Interlocked.CompareExchange` in `finally` clears `_currentLearnCts` only if it still points to our CTS, so disposal fires exactly once regardless of how many preemptions occurred.
- **E2E modal transitions (not `WaitForLearningMode`) to verify preemption:** `IsInLearningMode` stays `true` after the first command is cancelled (it's a one-way latch in the simulated device), so checking the modal message is the reliable signal that Power's learning actually started.


jodavis-claude (commented):
## GH-120 Code Review

Overall the cancel-and-wait design is correct and well-structured. One correctness issue was found (latent, currently unreachable) and one style nit.

src/AdaptiveRemote.App/Services/Broadlink/BroadlinkCommandService.cs
```diff
@@ -82,30 +88,33 @@ protected override Command.ExecuteDelegate CreateHandler(IRCommand command)
...
+            // Cancel any in-progress learning cycle, then wait for the semaphore it holds.
+            using CancellationTokenSource learnCts = new();
+            CancellationTokenSource? previousCts = Interlocked.Exchange(ref _currentLearnCts, learnCts);
+            if (previousCts is not null)
+            {
+                Logger.BroadlinkCommandService_CancellingLearningForNewCommand(command);
+                await previousCts.CancelAsync();
+            }
+
+            await _learningSemaphore.WaitAsync(cancellationToken);
```
jodavis-claude: **P1 — `_currentLearnCts` not cleared if `WaitAsync` throws**

`Interlocked.Exchange(ref _currentLearnCts, learnCts)` (line 97) sets `_currentLearnCts` to the new CTS before `WaitAsync` is called. If `WaitAsync` throws `OperationCanceledException` (outer CT cancelled before the semaphore is acquired), the lambda exits via exception, the `using` disposes `learnCts`, but the `try/finally` block is never entered — so `Interlocked.CompareExchange(ref _currentLearnCts, null, learnCts)` never runs. `_currentLearnCts` is left pointing to the now-disposed `learnCts`.

A subsequent `ProgramAsync` invocation would retrieve this disposed CTS via `Interlocked.Exchange`, then call `previousCts.CancelAsync()` on it — throwing `ObjectDisposedException` and faulting the new operation.

**Mitigating factor in current code**: The outer CT fires only when `_stop.Token` fires (service shutdown), which happens in `CleanUpAsync`. `CleanUpAsync` synchronously replaces all `ProgramAsync` handlers after `_stop.CancelAsync()`, so no new call to the old lambda can arrive on the Blazor dispatcher. This makes the bug unreachable today.

**Suggested fix** — wrap `WaitAsync` inside the `try` with a boolean guard:

```csharp
bool acquired = false;
try
{
    await _learningSemaphore.WaitAsync(cancellationToken);
    acquired = true;
    await _modalMessageService.ShowMessageAsync(...);
}
finally
{
    Interlocked.CompareExchange(ref _currentLearnCts, null, learnCts);
    if (acquired) _learningSemaphore.Release();
}
```

This ensures `_currentLearnCts` is always cleared on any exit path, without changing the observable behavior for current callers.

test/AdaptiveRemote.App.Tests/Services/Broadlink/BroadlinkCommandServiceTests.cs
```diff
@@ -477,10 +475,84 @@ public void BroadlinkCommandService_ProgramAsync_WhileLearningAlreadyInProgress_
...
+                    TaskCompletionSource tcs = new();
+                    ct.Register(() => tcs.TrySetCanceled(ct));
+                    return tcs.Task;
+                }
+                return Task.CompletedTask;
+            })
+            .Verifiable(Times.Exactly(2));
+
+        // Both calls use the same message; expect it to be shown twice
+        MockModalMessageService
```
jodavis-claude: nit: This test bypasses the `Expect_ModalMessageService_ShowMessage` helper and calls `MockModalMessageService.Setup()` directly. CONTRIBUTING.md requires wrapping all mock setup chains in `Expect_*` helpers.

The direct setup is necessary here because `Expect_ModalMessageService_ShowMessage` hard-codes `Times.Once`, and this scenario needs `Times.Exactly(2)` for the same message. Consider adding a `times` parameter overload to the helper:

```csharp
private void Expect_ModalMessageService_ShowMessage(string expectedMessage, Times? times = null)
    => MockModalMessageService
        .Setup(x => x.ShowMessageAsync(expectedMessage, It.IsAny<Func<CancellationToken, Task>>(), It.IsAny<bool>(), It.IsAny<CancellationToken>()))
        .Returns(delegate (string message, Func<CancellationToken, Task> action, bool keepAlive, CancellationToken ct)
        {
            Assert.AreEqual(expectedMessage, message);
            return action(ct);
        })
        .Verifiable(times ?? Times.Once);
```

Both new tests could then use the helper consistently.

src/AdaptiveRemote.App/Services/Broadlink/BroadlinkCommandService.cs
jodavis-claude: Fixed. Moved WaitAsync inside the try block with a bool acquired = false guard. CompareExchange now runs in finally on every exit path (including OperationCanceledException from WaitAsync before the semaphore is acquired), and Release() is conditioned on acquired. Style errors (missing braces, same-line statement) were corrected before committing.

test/AdaptiveRemote.App.Tests/Services/Broadlink/BroadlinkCommandServiceTests.cs
jodavis-claude: Fixed. Added an optional Times? times parameter (defaulting to Times.Once()) to the Expect_ModalMessageService_ShowMessage helper. Updated the DuplicateClick test to call Expect_ModalMessageService_ShowMessage(message, Times.Exactly(2)) and removed the raw MockModalMessageService.Setup() block.

src/AdaptiveRemote.App/Services/Broadlink/BroadlinkCommandService.cs
jodavis-claude: Confirmed fixed.  is now inside the  block (lines 107–108),  flag ensures  fires only when the semaphore was acquired, and  runs in  on every exit path including a  throw. P1 resolved.

test/AdaptiveRemote.App.Tests/Services/Broadlink/BroadlinkCommandServiceTests.cs
jodavis-claude: Confirmed fixed. The helper now accepts  (line 669) and the duplicate-click test calls  (line 516). No raw  calls remain. Nit resolved.

jodavis-claude (commented):
## GH-120 Sign-off Review

All prior review threads are resolved. No new issues were found in the three modified files (BroadlinkCommandService.cs, BroadlinkCommandServiceTests.cs, ServiceFixture.cs).

**Thread 1 — P1 correctness (BroadlinkCommandService.cs):** Resolved. WaitAsync is now inside the try block with an `acquired` boolean guard. CompareExchange fires in finally on every exit path, including OperationCanceledException from WaitAsync before semaphore acquisition. Release() is correctly conditioned on `acquired`.

**Thread 2 — Style nit (BroadlinkCommandServiceTests.cs):** Resolved. Expect_ModalMessageService_ShowMessage now accepts an optional `Times? times` parameter (defaulting to Times.Once()). The duplicate-click test uses the helper with Times.Exactly(2); no raw Setup calls remain.

**New file — ServiceFixture.cs:** Directory.Exists changed to Path.Exists for .git detection, correctly handling git worktrees where .git is a file rather than a directory. No issues.

Sign-off: **approved**.

jodavis-claude (commented):
## GH-120 Sign-off (commit 4be913d)

One commit was pushed after the previous sign-off (2026-05-23T02:49Z). This review covers that commit only.

**New commit — `4be913d` Fix race: HandleEnterLearningAsync cleared data queued by test**

`test/AdaptiveRemote.EndtoEndTests.TestServices/SimulatedBroadlink/SimulatedBroadlinkDevice.cs`

The removed line (`_pendingLearnedData = null`) inside the `HandleEnterLearningAsync` lock was redundant: `HandleCheckLearnedDataAsync` already clears `_pendingLearnedData` when it is consumed (line 438). The only effect of the cleared line was a race window where a test calling `ProvideLearnedData` before `HandleEnterLearningAsync` acquired the lock would have its queued data silently discarded, causing the poll loop to hang indefinitely. The fix is correct and the commit message accurately explains the reasoning.

No new issues found. All prior review threads remain resolved.

Sign-off: **approved**.

PR #203, opened 2026-05-30
ADR-221: Implement Sample class hierarchy and ManifestStore with JSON round-trip serialisation

jodavis-claude:
## Work item

ADR-221: Implement the `Sample` class hierarchy and `Manifest[S]` / `ManifestStore` in `ml/pipeline/core/` — the shared data model that every subsequent pipeline stage depends on, including JSON round-trip serialisation and the content-hash formula.

## Changes

- `ml/pipeline/__init__.py` — new empty package marker
- `ml/pipeline/core/__init__.py` — new empty package marker
- `ml/pipeline/core/sample.py` — all six dataclasses (`Sample`, `SampleWithPath`, `TextSample`, `AudioSample`, `SampleSpectrogram`, `SampleTokens`) with every field from the spec; `TextSample.id` assigned `str(uuid4())`
- `ml/pipeline/core/manifest.py` — `Manifest[S]` generic with O(1) `by_id`/`by_content_hash` index dicts; `ManifestStore.read()` (type-registry dispatch on `sample_type`) and `write()` (version-1 JSON)
- `ml/pyproject.toml` — minimal pytest config (`testpaths = ["test"]`, `pythonpath = ["."]`) enabling plain `pytest` invocation from `ml/`
- `ml/test/__init__.py` — new empty package marker
- `ml/test/pipeline/__init__.py` — new empty package marker
- `ml/test/pipeline/core/__init__.py` — new empty package marker
- `ml/test/pipeline/core/test_manifest.py` — 26 unit tests covering all four round-trip scenarios, `sample_type` dispatch, `TextSample.seed == 0` serialisation, `SampleSpectrogram.parent_id` preservation, and `AudioSample.applied_values` numeric-type fidelity

## Design decisions

- **`seed` has no dataclass default** — the spec states "PhraseVariator explicitly sets `seed=0` at construction", meaning callers pass it explicitly; a default would silently mask missing arguments in later stages.
- **Mutable dataclasses (`frozen=False`)** — the spec is silent on `frozen=`; downstream `ModifierStage` uses `Sample` instances as dict *values* (not keys), so mutability is safe and avoids unnecessary copy overhead.
- **`pyproject.toml` created in `ml/`** — resolves the known ambiguity that `validate-build` and `validate-tests` only cover .NET; enables `pytest ml/test/pipeline/core/` without manual `PYTHONPATH` manipulation.

---
_Generated by [Claude Code](https://claude.ai/code/session_01DLVaMxWuhm75L3YaMdVTFT)_

### DIFF
--- a/ml/pipeline/core/sample.py
+++ b/ml/pipeline/core/sample.py
@@ -1,0 +1,72 @@
+from __future__ import annotations
+
+from abc import ABC
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any
+
+
+@dataclass
+class Sample(ABC):
+    """Base class for all pipeline sample types.
+
+    id: stable identifier.
+      TextSample: str(uuid4()) — no file, not user-visible.
+      SampleWithPath subtypes: derived by each stage's _derive_id(); used as filename stem.
+    seed: 0 for TextSample and deterministic stages; os.urandom(8) for stochastic stages.
+    content_hash: sha256 hex digest.
+      TextSample: sha256(content.encode('utf-8')).
+      SampleWithPath subtypes: sha256(parent_content_hash + ":" + str(seed) + ":" + canonical(applied_values)).
+    """
+
+    id: str
+    seed: int
+    content_hash: str
+
+
+@dataclass
+class SampleWithPath(Sample, ABC):
+    """Base for all ModifierStage output types that produce a file.
+
+    path: relative filename (bare name only, no directory component).
+    parent_content_hash: content_hash of the upstream sample; skip-unchanged lookup key.
+    """
+
+    path: Path
+    parent_content_hash: str
+
+
+@dataclass
+class TextSample(Sample):
+    """Bootstrapped phrase variant; no output file.
+
+    Callers set id=str(uuid4()) and seed=0.
+    content_hash = sha256(content.encode('utf-8')).
+    """
+
+    content: str
+    label: str
+
+
+@dataclass
+class AudioSample(SampleWithPath):
+    """WAV file produced by TtsSampleGenerator or an augmentation stage."""
+
+    transcript: str
+    applied_values: dict[str, Any]
+
+
+@dataclass
+class SampleSpectrogram(SampleWithPath):
+    """NPY file produced by SpectrogramStage (seed=0, deterministic)."""
+
+    transcript: str
+    parent_id: str
+
+
+@dataclass
+class SampleTokens(SampleWithPath):
+    """JSON file produced by TokenStage (seed=0, deterministic)."""
+
+    transcript: str
+    parent_id: str
--- a/ml/pipeline/core/sample.py
+++ b/ml/pipeline/core/sample.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 from abc import ABC
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from pathlib import Path
 from typing import Any
 
@@ -11,7 +11,7 @@ class Sample(ABC):
     """Base class for all pipeline sample types.
 
     id: stable identifier.
-      TextSample: str(uuid4()) — no file, not user-visible.
+      TextSample: equals content_hash — content-addressable; no file, not user-visible.
       SampleWithPath subtypes: derived by each stage's _derive_id(); used as filename stem.
     seed: 0 for TextSample and deterministic stages; os.urandom(8) for stochastic stages.
     content_hash: sha256 hex digest.
@@ -40,12 +40,17 @@ class SampleWithPath(Sample, ABC):
 class TextSample(Sample):
     """Bootstrapped phrase variant; no output file.
 
-    Callers set id=str(uuid4()) and seed=0.
+    id is derived from content_hash (content-addressable; callers do not pass it).
+    seed=0 for all TextSamples.
     content_hash = sha256(content.encode('utf-8')).
     """
 
     content: str
     label: str
+    id: str = field(init=False)
+
+    def __post_init__(self) -> None:
+        self.id = self.content_hash
 
 
 @dataclass
--- a/ml/pipeline/core/manifest.py
+++ b/ml/pipeline/core/manifest.py
@@ -148,7 +148,6 @@ def _serialise(s: Sample) -> dict[str, Any]:
 def _deserialise(cls: type, e: dict[str, Any]) -> Sample:
     if cls is TextSample:
         return TextSample(
-            id=e["id"],
             seed=e["seed"],
             content_hash=e["content_hash"],
             content=e["content"],
--- a/ml/pipeline/core/manifest.py
+++ b/ml/pipeline/core/manifest.py
@@ -1,0 +1,162 @@
+from __future__ import annotations
+
+import json
+from pathlib import Path
+from typing import Any, Generic, Sequence, TypeVar
+
+from pipeline.core.sample import (
+    AudioSample,
+    Sample,
+    SampleSpectrogram,
+    SampleTokens,
+    TextSample,
+)
+
+S = TypeVar("S", bound=Sample)
+
+
+class Manifest(Generic[S]):
+    """Typed, immutable collection of samples with O(1) lookup by id or content_hash."""
+
+    def __init__(self, samples: Sequence[S]) -> None:
+        self._samples: tuple[S, ...] = tuple(samples)
+        self._by_content_hash: dict[str, S] = {s.content_hash: s for s in self._samples}
+        self._by_id: dict[str, S] = {s.id: s for s in self._samples}
+
+    @property
+    def samples(self) -> tuple[S, ...]:
+        return self._samples
+
+    def by_content_hash(self, h: str) -> S | None:
+        return self._by_content_hash.get(h)
+
+    def by_id(self, id: str) -> S | None:
+        return self._by_id.get(id)
+
+
+_SAMPLE_TYPE_KEY: dict[type, str] = {
+    TextSample: "text",
+    AudioSample: "audio",
+    SampleSpectrogram: "spectrogram",
+    SampleTokens: "tokens",
+}
+
+_SAMPLE_TYPE_CLASS: dict[str, type] = {v: k for k, v in _SAMPLE_TYPE_KEY.items()}
+
+
+class ManifestStore:
+    """Reads and writes Manifest JSON files (schema version 1).
+
+    JSON format:
+      {"version": 1, "sample_type": "<type>", "samples": [...]}
+
+    Path fields in JSON are bare filenames (no directory component).
+    Callers prepend output_dir when resolving full paths.
+    Numeric applied_values are stored as raw int/float — never as strings.
+    """
+
+    def read(self, path: Path) -> Manifest[Any]:
+        with open(path, "r", encoding="utf-8") as f:
+            data = json.load(f)
+        cls = _SAMPLE_TYPE_CLASS[data["sample_type"]]
+        samples = [_deserialise(cls, entry) for entry in data["samples"]]
+        return Manifest(samples)
+
+    def write(self, manifest: Manifest[Any], path: Path) -> None:
+        samples = manifest.samples
+        sample_type = _SAMPLE_TYPE_KEY[type(samples[0])]
+        payload: dict[str, Any] = {
+            "version": 1,
+            "sample_type": sample_type,
+            "samples": [_serialise(s) for s in samples],
+        }
+        with open(path, "w", encoding="utf-8") as f:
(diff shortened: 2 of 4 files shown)


jodavis-claude (commented):
**Priority 1 — Correctness**

- **ml/pipeline/core/manifest.py line 67:** `ManifestStore.write()` raises `IndexError` on an empty manifest. `sample_type = _SAMPLE_TYPE_KEY[type(samples[0])]` throws when `manifest.samples` is empty. `Manifest([])` is a valid object (see `test_empty_manifest`), so `write()` must either support it or raise a descriptive `ValueError("Cannot write empty manifest")` instead of an opaque index error.

- **ml/pipeline/core/manifest.py line 61:** `ManifestStore.read()` ignores the `"version"` field entirely. The JSON schema explicitly carries `"version": 1`. Without a version check, a file written by a future incompatible schema is silently deserialized as version 1 and produces corrupt objects with no diagnostic. Add `if data.get("version") != 1: raise ValueError(f"Unsupported manifest version: {data.get('version')!r}")`.

**Priority 4 — Test coverage**

- **ml/test/pipeline/core/test_manifest.py:** No test exercises `ManifestStore().write(Manifest([]), path)`. The existing `test_empty_manifest` creates `Manifest([])` and calls lookup methods but never calls `write()`, so the `IndexError` from issue 1 is not caught by the test suite. Add a test under `TestManifestLookup` (or a dedicated class) that confirms the behaviour of writing an empty manifest.

**Style (non-blocking)**

- **ml/test/pipeline/core/test_manifest.py line 4:** `import tempfile` is unused — tests use the `tmp_path` pytest fixture.
- **ml/pipeline/core/manifest.py line 61:** `_SAMPLE_TYPE_CLASS[data["sample_type"]]` raises a bare `KeyError` on unrecognised types. A `ValueError(f"Unknown sample_type: {data['sample_type']!r}")` would be more informative.

---
_Generated by [Claude Code](https://claude.ai/code)_

.claude/agents/debugger.md
```diff
@@ -14,6 +14,8 @@ tools:
   - Bash
   - Task
   - Skill
+  - mcp__github__issue_read
+  - mcp__github__add_issue_comment
```
jodavis: Remove these, they're no longer necessary

.claude/agents/developer.md
```diff
@@ -21,6 +21,7 @@ tools:
   - mcp__jira__transitionJiraIssue
   - mcp__jira__editJiraIssue
   - mcp__jira__addCommentToJiraIssue
+  - mcp__github__create_pull_request
```
jodavis: Remove, no longer necessary

.claude/agents/researcher.md
```diff
@@ -16,6 +16,7 @@ tools:
   - mcp__microsoft-learn__microsoft_docs_search
   - mcp__microsoft-learn__microsoft_docs_fetch
   - mcp__microsoft-learn__microsoft_code_sample_search
+  - mcp__github__issue_read
```
jodavis: Remove

.claude/agents/reviewer.md
```diff
@@ -12,6 +12,9 @@ tools:
   - Grep
   - Bash
   - Skill
+  - mcp__github__pull_request_read
+  - mcp__github__pull_request_review_write
+  - mcp__github__update_pull_request
```
jodavis: Remove

.claude/commands/debugger-investigate.md
```diff
@@ -18,12 +18,17 @@ If missing, stop and tell the caller:
 ### 1 — Fetch the GitHub issue
 
 Derive the issue number by stripping the `Issue-` prefix from the work item ID
-(e.g. `Issue-444` → `444`). Fetch the issue title, body, and all comments:
+(e.g. `Issue-444` → `444`). Fetch the issue title, body, and all comments. Prefer `gh`
```
jodavis: The top level session will have to do this using the GitHub MCP.

.claude/commands/dev-team.md
```diff
@@ -1,5 +1,5 @@
 ---
-description: Entry point for the dev-team agent pipeline. Routes to the correct pipeline based on the work item request, then starts the pipeline script and relays its output.
+description: Entry point for the dev-team agent pipeline. Routes to the correct pipeline based on the work item request, starts the pipeline script, and acts as the active orchestrator — handling GitHub and Jira operations on behalf of the pipeline when requested.
```
jodavis: The top level agent isn't the orchestrator, but it is responsible for reporting all progress. Think of it as the scrum master, getting updates from team members and updating work items, GitHub PRs and issues. 

At a high level, the orchestrator directs the team members but doesn't tell the scrum master what to do. The  scrum master just listens to progress reports and updates things appropriately.

.claude/commands/dev-team.md
```diff
@@ -9,19 +9,21 @@ $ARGUMENTS
 
 ## Role
 
-You represent a development team. Your job is to take a work item request, choose a work 
-plan (pipeline), and then hand off the work item and the plan the pipeline script. Then you
-report progress to the user by watching the output of the script. The script is the orchestrator.
-Once you have started the script you are a passive observer.
+You are the active orchestrator for the dev-team pipeline. You start the pipeline script,
```
jodavis: Not the active orchestrator

.claude/commands/dev-team.md
```diff
@@ -61,14 +63,56 @@ python -c "import sys; print(sys.platform)"
...
+   | `create_pr` | `mcp__github__create_pull_request` — use `owner`, `repo`, `title`, `head`, `base`, `body`, `draft: true` from the action; result contains the PR URL |
+   | `post_review` | `mcp__github__pull_request_review_write` — `method: "create"`, `event: "COMMENT"`, `owner`, `repo`, `pullNumber`, `body` |
+   | `get_review_threads` | `mcp__github__pull_request_read` — `method: "get_review_comments"`, `owner`, `repo`, `pullNumber`; return the full threads array |
+   | `resolve_threads` | For each `thread_id` in `thread_ids`: `mcp__github__pull_request_review_write` with `method: "resolve_thread"`, `threadId`, `owner`, `repo`, `pullNumber` |
+   | `pr_ready` | `mcp__github__update_pull_request` — `draft: false`, `owner`, `repo`, `pullNumber` |
+   | `add_reviewer` | `mcp__github__update_pull_request` — `reviewers: [reviewer]`, `owner`, `repo`, `pullNumber` |
+   | `jira_comment` | `mcp__jira__addCommentToJiraIssue` — `issue_id`, `body` from the action |
+   | `jira_transition` | `mcp__jira__transitionJiraIssue` — `issue_id`, `transition` from the action |
+
+3. **Write** the result to:
```
jodavis: Do not write the results to the context file. Since the team members aren't accessing the PR directly anymore, they don't need the PR URL. (It would be nice to report that to the user though). None of the other results should matter either.

.claude/commands/dev-team.md
```diff
@@ -61,14 +63,56 @@ python -c "import sys; print(sys.platform)"
...
 - **anything else**: `tail -f <task-output-path>`
 
-Stream all output to the user as it arrives until the process exits.
+Relay each notification to the user as it arrives. When you see a line matching
+`@@ACTION_REQUIRED: <type>`, pause relaying and handle the action (see Step 4a). Resume
+monitoring afterward. Continue until the process exits.
+
+### 4a — Handle action requests
+
+When `@@ACTION_REQUIRED: <type>` appears in the pipeline output:
```
jodavis: Instead of looking for this token, the top level agent will just watch output from the thread and have its own rules about what to do when certain things happen. 

To get the details it needs, for things like the PR description or comment details, it can read the context file.

.claude/commands/prepare-pr-details.md
jodavis, line 1: Rename this from "developer-create-or" since it's not the developers skill anymore.

jodavis (changes requested):
This review is focused only on the dev-team changes. Make these changes, and overall the design should be that subagents work together, with the script as the orchestrator, but they don't interact directly with GitHub or Jira. The top level agent passively watches the script's progress and makes all the updates to PRs, work items, and issues. If there's subs additional information that the top level agent needs to complete an update, we should find a way to plumb it through. But in general, the status updates front the script and the content in the context file should be enough. 

There is one other thing: since the developer and reviewer can no longer communicate through PR threads, they will have to pass a structured list of comment threads back and forth, including all the comments and their responses, who they came from, and whether or not there are resolved. That structure with have to go into the context file as well, so the top level session can see them. Might need to add an ID as well, so the top level agent can map them to GitHub PR comment threads and update the right ones. 

Also, there is a bug again that the review comments were added as a normal comment instead of a review. All the rules about PR comments that we added in the reviewer need to be transferred to the top level agent now.

jodavis-claude (commented):
No prior review threads.

Two new Priority 1 (correctness/fault tolerance) issues found in ml/pipeline/core/manifest.py:

1. manifest.py:63 — data["sample_type"] is an unguarded dict subscript. The preceding version check uses data.get("version") defensively, but sample_type access raises KeyError instead of ValueError for a malformed file. Use data.get("sample_type") and raise ValueError explicitly to match the established pattern.

2. manifest.py:70-81 — write() does not validate that all samples share the same type. Python generics are erased at runtime, so Manifest([TextSample(...), AudioSample(...)]) is constructable. If passed to write(), the method silently produces a JSON file whose sample_type header matches only the first entry; on read(), _deserialise tries to construct the wrong class for the remaining entries and fails with a confusing KeyError far from the original mistake — and the corrupt file is already on disk. Add a guard at the top of write(): if any(type(s) is not type(samples[0]) for s in samples): raise ValueError(f"Manifest contains mixed sample types: {set(type(s).__name__ for s in samples)}").

All exit criteria are otherwise satisfied: all six dataclasses are present with fields matching spec lines 272-309; Manifest[S] with O(1) by_content_hash/by_id; ManifestStore read/write with discriminated dispatch; JSON schema version 1; all four round-trip tests plus dispatch, seed=0, parent_id, and numeric-type tests. The recent fix commits (version validation, ValueError for unknown sample_type, empty-manifest guard) are all correct additions.

---
_Generated by [Claude Code](https://claude.ai/code)_

.claude/agents/debugger.md
jodavis: This change is no longer part of this PR.

.claude/agents/developer.md
jodavis: This change is no longer part of this PR.

.claude/agents/reviewer.md
jodavis: This change is no longer part of this PR.

.claude/agents/researcher.md
jodavis: This change is no longer part of this PR.

.claude/commands/debugger-investigate.md
jodavis: This change is no longer part of this PR.

.claude/commands/dev-team.md
jodavis: This change is no longer part of this PR.

.claude/commands/dev-team.md
jodavis: This change is no longer part of this PR.

.claude/commands/dev-team.md
jodavis: This change is no longer part of this PR.

.claude/commands/dev-team.md
jodavis: This change is no longer part of this PR.

.claude/commands/prepare-pr-details.md
jodavis: This change is no longer part of this PR.

ml/pipeline/core/sample.py
```diff
@@ -0,0 +1,72 @@
...
+    id: stable identifier.
+      TextSample: str(uuid4()) — no file, not user-visible.
+      SampleWithPath subtypes: derived by each stage's _derive_id(); used as filename stem.
+    seed: 0 for TextSample and deterministic stages; os.urandom(8) for stochastic stages.
+    content_hash: sha256 hex digest.
+      TextSample: sha256(content.encode('utf-8')).
+      SampleWithPath subtypes: sha256(parent_content_hash + ":" + str(seed) + ":" + canonical(applied_values)).
+    """
+
+    id: str
```
jodavis, line 22: Could we just use the hash as the ID? Ideally we shouldn't have samples with the same hash, because that means we have duplicate samples.

jodavis (changes requested):

jodavis:
@copilot Address all PR review comments

jodavis (approved):

PR #214, opened 2026-06-02
ADR-222: Implement seed-based randomisation engine

jodavis-claude:
## Work item: ADR-222

Implement the seed-based randomisation engine (`PassFilter`, `MinMaxFilter`, `NormalFilter`, `VariationGenerator`) — the foundational component that all `ModifierStage` subclasses will call to produce deterministic, stable applied values for each sample.

## Changes

- `ml/pipeline/core/randomization.py` — New file. Implements `PassFilter` ABC, `MinMaxFilter` (uniform over `[min_val, max_val]`), `NormalFilter` (Gaussian with peak-normalised density), and `VariationGenerator` with `should_vary`, `generate`, `generate_int`, and `choose`. Stdlib-only (`hashlib`, `math`, `abc`, `typing`); no external dependencies. Input validation added: `MinMaxFilter` raises `ValueError` when `min_val > max_val`; `should_vary` raises `ValueError` when `frequency` is outside `[0.0, 1.0]`; `choose` raises `ValueError` when `options` is empty.
- `ml/test/pipeline/core/test_randomization.py` — New file. 56 unit tests across 8 test classes covering all control flow branches, error paths, boundary inputs, and the determinism/stability contract.
- `scripts/validate-tests.sh` — Python pytest run added (`cd ml || exit 1 && python3 -m pytest`).
- `scripts/validate-tests.cmd` — Python pytest run added (`cd ml` with `%ERRORLEVEL%` check, then `python -m pytest`).
- `.github/workflows/build-and-test.yml` — New `python-tests` job on `ubuntu-latest`: checks out, sets up Python 3.12, installs `pytest`, and runs `python -m pytest --verbosity=1` from `ml/`. Only `pytest` is installed; no heavy ML dependencies required since all `pipeline` modules are stdlib-only.

## Design decisions

- **Stability tests use `generate_int`, not `generate`**: Float `generate` with `MinMaxFilter` always scales proportionally when the domain changes (no genuinely stable seeds possible). Bitmask rejection in `generate_int` produces real stability for seeds where the raw hash falls below the original mask threshold. Seeds 1, 2, and 5 were selected by an offline sweep over seeds 0–199.
- **`_NeverAcceptFilter` helper in test file**: Used to trigger the 1000-iteration `ValueError` guard in `generate` without requiring a real filter with pathological inputs.
- **`generate_int` reads bounds via `sample_domain()`**: Uses the public interface rather than accessing private attributes on the filter.
- **Python CI job is separate from the .NET job**: The Python tests have no .NET dependencies and run faster on `ubuntu-latest`, keeping the jobs independent and parallelised.

### DIFF
--- a/ml/pipeline/core/randomization.py
+++ b/ml/pipeline/core/randomization.py
@@ -16,6 +16,20 @@ def _hash_int(key: str) -> int:
 
 
 class PassFilter(ABC):
+    def __init__(self, precision: int = 0) -> None:
+        domain_low, domain_high = self.sample_domain()
+        scale = 10 ** precision
+        low_s = round(domain_low * scale)
+        high_s = round(domain_high * scale)
+        bias_s = max(0, -low_s)
+        shifted_high = high_s + bias_s
+        self._precision = precision
+        self._scale = scale
+        self._low_s = low_s
+        self._high_s = high_s
+        self._bias_s = bias_s
+        self._pow2_range = 1 << math.ceil(math.log2(shifted_high + 1)) if shifted_high > 0 else 1
+
     @abstractmethod
     def density(self, value: float) -> float:
         """Normalised density; max == 1.0. Acceptance probability in rejection sampling."""
@@ -30,13 +44,14 @@ class PassFilter(ABC):
 class MinMaxFilter(PassFilter):
     """Uniform over [min_val, max_val]. density() == 1.0 in range, 0.0 outside."""
 
-    def __init__(self, min_val: float, max_val: float) -> None:
+    def __init__(self, min_val: float, max_val: float, *, precision: int = 0) -> None:
         if min_val > max_val:
             raise ValueError(
                 f"min_val must be <= max_val, got min_val={min_val}, max_val={max_val}"
             )
         self._min_val = min_val
         self._max_val = max_val
+        super().__init__(precision)
 
     def density(self, value: float) -> float:
         return 1.0 if self._min_val <= value <= self._max_val else 0.0
@@ -48,11 +63,12 @@ class MinMaxFilter(PassFilter):
 class NormalFilter(PassFilter):
     """Gaussian. density(x) = gaussian_pdf(x)/gaussian_pdf(mean); peak == 1.0."""
 
-    def __init__(self, mean: float, std_dev: float) -> None:
+    def __init__(self, mean: float, std_dev: float, *, precision: int = 0) -> None:
         if std_dev <= 0:
             raise ValueError(f"std_dev must be positive, got {std_dev}")
         self._mean = mean
         self._std_dev = std_dev
+        super().__init__(precision)
 
     def density(self, value: float) -> float:
         z = (value - self._mean) / self._std_dev
@@ -74,12 +90,15 @@ class VariationGenerator:
         return (raw / _TWO_TO_64) < frequency
 
     def generate(self, variable_name: str, pass_filter: PassFilter) -> float:
-        """Rejection-sample deterministically using attempt-indexed sha256 hashes."""
-        domain_low, domain_high = pass_filter.sample_domain()
-        domain_width = domain_high - domain_low
+        """Rejection-sample using power-of-2 modulo for stability across domain changes."""
+        if pass_filter._high_s == pass_filter._low_s:
+            return pass_filter._low_s / pass_filter._scale
+        pow2_range = pass_filter._pow2_range
+        bias_s = pass_filter._bias_s
+        scale = pass_filter._scale
         for n in range(_MAX_ATTEMPTS):
             raw = _hash_int(f"{self._seed}:{variable_name}:{n}")
-            candidate = domain_low + (raw / _TWO_TO_64) * domain_width
+            candidate = (raw % pow2_range - bias_s) / scale
             accept_raw = _hash_int(f"{self._seed}:{variable_name}:{n}:accept")
             if (accept_raw / _TWO_TO_64) < pass_filter.density(candidate):
                 return candidate
--- a/ml/pipeline/core/randomization.py
+++ b/ml/pipeline/core/randomization.py
@@ -31,6 +31,10 @@ class MinMaxFilter(PassFilter):
     """Uniform over [min_val, max_val]. density() == 1.0 in range, 0.0 outside."""
 
     def __init__(self, min_val: float, max_val: float) -> None:
+        if min_val > max_val:
+            raise ValueError(
+                f"min_val must be <= max_val, got min_val={min_val}, max_val={max_val}"
+            )
         self._min_val = min_val
         self._max_val = max_val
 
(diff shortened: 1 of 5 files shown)


ml/test/pipeline/core/test_randomization.py
```diff
@@ -0,0 +1,332 @@
...
+        true_count = sum(
+            1 for s in range(n_seeds)
+            if VariationGenerator(s).should_vary("v", frequency)
+        )
+        ratio = true_count / n_seeds
+        # Expect within 8% of the target frequency for 1000 samples
+        assert abs(ratio - frequency) < 0.08
+
+    def test_different_variable_names_are_independent(self) -> None:
+        g = VariationGenerator(0)
```
jodavis-claude: Style (not blocking): Asserts only type, not independence. `TestVariationGeneratorGenerate::test_different_variable_names_produce_independent_values` uses `va != vb` which is stronger. Seed=0 does give different results for `prefix_delay_s` vs `suffix_delay_s` — consider asserting that directly.

ml/test/pipeline/core/test_randomization.py
```diff
@@ -0,0 +1,332 @@
...
+    def test_stable_seed_same_value_when_max_widened(self) -> None:
+        # seed=1: value 2 is below 10 and its raw bits have bit-4 == 0, so mask=31 gives
+        # the same 2 as mask=15.
+        v_narrow = VariationGenerator(1).generate_int("x", MinMaxFilter(0, 10))
+        v_wide = VariationGenerator(1).generate_int("x", MinMaxFilter(0, 20))
+        assert v_narrow == 2
+        assert v_wide == 2
+
+    def test_changing_seed_gets_higher_when_max_widened(self) -> None:
+        # seed=2: raw_0 & 15 == 3 (accepted for max=10), raw_0 & 31 == 11 (accepted for max=20).
```
jodavis-claude: Style (not blocking): `raw_0 & 15 == 3` and `raw_0 & 31 == 11` cannot both be true for the same raw_0: lower 4 bits of 0011 (=3) means lower 5 bits are either 00011 (=3) or 10011 (=19), not 01011 (=11). The asserted values are correct (verified offline), but the comment explaining the mechanism is wrong. Most likely seed=2 loops to n>0 for one of the two masks, meaning they draw from different hash keys.

ml/test/pipeline/core/test_randomization.py
```diff
@@ -0,0 +1,332 @@
...
+        assert VariationGenerator(seed).choose("voice", options) == expected
+
+    def test_different_seeds_can_choose_different_items(self) -> None:
+        options = ["en-US-JennyNeural", "en-US-GuyNeural", "en-GB-LibbyNeural"]
+        # seed=0 -> JennyNeural, seed=1 -> GuyNeural
+        assert VariationGenerator(0).choose("voice", options) == "en-US-JennyNeural"
+        assert VariationGenerator(1).choose("voice", options) == "en-US-GuyNeural"
+
+    def test_different_variable_names_select_independently(self) -> None:
+        options = ["x", "y", "z"]
```
jodavis-claude, line 404: Style (not blocking): Same issue as `TestVariationGeneratorShouldVary::test_different_variable_names_are_independent` — asserts only type. Seed=0 selects different options for `noise_file` vs `voice`; asserting `a != b` would be more definitive.

jodavis-claude (commented):
All exit criteria for ADR-222 are met. No Priority 1-4 issues found.

Two style-level observations (not blocking):

1. `test_different_variable_names_are_independent` (TestVariationGeneratorShouldVary) and `test_different_variable_names_select_independently` (TestVariationGeneratorChoose) only assert return type. The `TestVariationGeneratorGenerate` class does this better with `test_different_variable_names_produce_independent_values` (asserts `va != vb`). Consider applying the same pattern.

2. The comment in `TestVariationGeneratorGenerateIntStability` is internally inconsistent: it claims `raw_0 & 15 == 3` and `raw_0 & 31 == 11` for the same hash. Lower 4 bits of 0011 (=3) means lower 5 bits can only be 00011 (=3) or 10011 (=19), not 01011 (=11). The asserted values are correct, but the explanation of why they differ is wrong.

ml/pipeline/core/randomization.py
```diff
@@ -0,0 +1,107 @@
...
+        raw = _hash_int(f"{self._seed}:{variable_name}:vary")
+        return (raw / _TWO_TO_64) < frequency
+
+    def generate(self, variable_name: str, pass_filter: PassFilter) -> float:
+        """Rejection-sample deterministically using attempt-indexed sha256 hashes."""
+        domain_low, domain_high = pass_filter.sample_domain()
+        domain_width = domain_high - domain_low
+        for n in range(_MAX_ATTEMPTS):
+            raw = _hash_int(f"{self._seed}:{variable_name}:{n}")
+            candidate = domain_low + (raw / _TWO_TO_64) * domain_width
```
jodavis: This is not the design for how the candidate should be computed. The problem is that if the min or max changes the value will definitely change, because the domain_width or domain_low have changed.

Instead, the algorithm design should be this:

1. raw = _hash_int as today
2.  Compute the d=power-of-2 that contains the domain
3. candidate = raw % d
4. pass_filter.density(candidate) may still be 0 for min/max because the power-of-2 range is larger than the domain

As an example, suppose min/max is [0, 5], then d=8 (smallest power-of-2 > 5)
| raw | d | candidate | accepted |
| --- | --- | --- | --- |
| 12 | 8 | 12 % 8 = 4 | 4 |
| 6 | 8 | 6 % 8 = 6 | try again |
| 4 | 8 | 4 % 8 = 4 | 4 |
| 31 | 8 | 31 % 8 = 7 | try again |
| 20 | 8 | 20 % 8 = 4 | 4 |

Now, suppose the min/max changes to [0, 15], then d=16 (smallest power-of-2 > 15)
| raw | d | candidate | accepted | change |
| --- | --- | --- | --- | --- |
| 12 | 16 | 12 % 16 = 12 | 12 | increased because of larger domain |
| 6 | 16 | 6 % 16 = 6 | 6 | accepted instead of trying again |
| 4 | 16 | 4 % 16 = 4 | 4 | stable |
| 31 | 16 | 31 % 16 = 15 | 15 | accepted instead of trying again |
| 20 | 16 | 20 % 16 = 4 | 4 | stable |

So you see, as the range changes, the third and fifth values remained the same, while a proportion of the values changed in response to the increased range. This will allow us to keep some samples that are still valid in the new range, while changing a proportion of the samples depending on how large the change is.

.claude/settings.json
```diff
@@ -1,4 +1,11 @@
 {
+  "enabledPlugins": {
+    "dotnet@dotnet-agent-skills": false,
+    "dotnet-diag@dotnet-agent-skills": true,
+    "dotnet-aspnet@dotnet-agent-skills": false,
+    "dotnet-template-engine@dotnet-agent-skills": false,
+    "dev-team@dev-team-agents": true
+  },
```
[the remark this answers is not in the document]
jodavis: The PR will be scoped to the randomization engine, I was just using the new skill plug-in here to test it.

jodavis:
@copilot address all code review comments.

jodavis:
@copilot we're starting to build up a comprehensive set of Python unit tests in this project. Add them to the CI runs and to the `scripts/validate-tests.*` scripts.

jodavis (approved):

ml/pipeline/core/randomization.py
jodavis: This was resolved, the algorithm now uses a power-of-2 domain to keep some values stable while allowing others to vary.

.claude/settings.json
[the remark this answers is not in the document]
jodavis: The PR is now scoped to include only the randomization engine changes.

PR #218, opened 2026-06-11
ADR-278: ADR-191  ADR-278    1      OOP Pipeline: DVC wiring for intent stages

jodavis:
## ADR-278: Wire intent stages into DVC pipeline

Closes ADR-278.

### Summary

Wires the two OOP intent-stage entry-points (`intent_01_generate_phrases.py` and `intent_02_compute_vocab.py`) into DVC, completing Task 3b of the OOP ML pipeline feature.

### Changes

| File | Action |
|------|--------|
| `ml/params.yaml` | Added `pipeline.input_phrases_path: scripts/intent_prediction/01_input_phrases.csv` |
| `ml/download_phoneme_dictionary.py` | Created — downloads `cmudict-0.7b` from CMU Sphinx SVN to `data/phoneme_dictionary/cmudict-0.7b` |
| `ml/dvc.yaml` | Created — defines three stages: `download_phoneme_dictionary`, `generate_phrases`, `compute_vocab` |
| `ml/.dvc/config` | Created — sets `adr-ml-training-data` as the default S3 remote (`s3://adr-ml-training-data/dvc`) |
| `ml/.dvc/.gitignore` | Created — ignores `/tmp` and `/cache` inside `.dvc/` |
| `ml/.gitignore` | Created — ignores `.dvc/tmp`, `.dvc/cache`, and `/data/` |

### Key Decisions

- **DVC subdirectory init:** `.dvc/config` and `.dvc/.gitignore` written directly; `dvc init --subdir` can be run once from `ml/` if DVC reports missing internals.
- **`cmd` hardcodes input phrases path:** DVC `${}` interpolation in `cmd` only works in `foreach` stages; path is hardcoded in `cmd` and tracked via `params:` so DVC re-runs if the param changes.
- **`download_phoneme_dictionary.py` written from scratch:** stdlib-only (`urllib.request.urlretrieve`); no prior version found in git history.
- **`persist: true` not used:** Neither intent stage uses `ModifierStage`.

### Validation

`validate-build` and `validate-tests` pass — no C# or Python unit test changes.

`dvc repro` (cwd `ml/`) will: download CMU dict → generate phrase variants → compute vocab.


ml/.dvc/.gitignore
jodavis, line 1: Do we need this .gitignore file if the one a level up includes both of these paths? If not, delete this one.

ml/download_phoneme_dictionary.py
```diff
@@ -0,0 +1,27 @@
...
+Output: data/phoneme_dictionary/cmudict-0.7b
+"""
+
+from __future__ import annotations
+
+import urllib.request
+from pathlib import Path
+
+_URL = "http://svn.code.sf.net/p/cmusphinx/code/trunk/cmudict/cmudict-0.7b"
+_OUTPUT_DIR = Path(__file__).parent / "data" / "phoneme_dictionary"
```
jodavis: output dir should be an argument passed into the script, like all the other stage scripts

ml/download_phoneme_dictionary.py
```diff
@@ -0,0 +1,27 @@
+"""Download the CMU Pronouncing Dictionary (cmudict-0.7b) into data/phoneme_dictionary/.
```
jodavis: This file shouldn't be in the root ml/ folder. It should be under `ml/pipeline` like other scripts.

ml/dvc.yaml
```diff
@@ -0,0 +1,38 @@
+stages:
+  download_phoneme_dictionary:
+    cmd: python download_phoneme_dictionary.py
+    outs:
+      - data/phoneme_dictionary/
+
+  generate_phrases:
```
jodavis: prefix stages with `intent_NN_` to match the scripts. It will make it easier to find things as the pipeline grows

PR #231, opened 2026-06-18
ADR-227: Add BackgroundNoiseAugmentor and MicrophoneNoiseAugmentor stages

jodavis-claude:
**Work item:** ADR-227 — Implement `BackgroundNoiseAugmentor` and `MicrophoneNoiseAugmentor`, two `ModifierStage` subclasses that add environmental noise and microphone noise to `AudioSample` manifests, along with their DVC entry-point scripts.

## Changes

- `ml/pipeline/speech/background_noise_stage.py` — New file: `NoiseProvider` protocol and `BackgroundNoiseAugmentor(ModifierStage[AudioSample, AudioSample])`. `choose()` always called for hash stability; all three applied_values keys always present; `noise_start_s` bounds derived from runtime file durations via `AudioReader`.
- `ml/pipeline/speech/mic_noise_stage.py` — New file: `MicrophoneNoiseAugmentor(ModifierStage[AudioSample, AudioSample])`. Generates Gaussian noise seeded from `output_seed`. Single applied_values key: `mic_noise_amplitude`.
- `ml/pipeline/stages/speech_05_add_background_noise.py` — New entry-point: `--input-manifest-dir`, `--noise-dir`, `--output-dir`; constructs `BackgroundNoiseAugmentor` with `_DirectoryNoiseProvider`.
- `ml/pipeline/stages/speech_06_add_mic_noise.py` — New entry-point: `--input-manifest-dir`, `--output-dir`; constructs `MicrophoneNoiseAugmentor`.
- `ml/pipeline/stages/params.py` — Added `AddBackgroundNoiseParams` and `AddMicNoiseParams` dataclasses; extended `PipelineParams` and `PipelineParams.load()` to parse the new sections.
- `ml/params.yaml` — Added `stages.add_background_noise` and `stages.add_mic_noise` sections with default values.
- `ml/pipeline/speech/_doc_speech.md` — Updated stages table to include both new stages; added design decisions for `NoiseProvider`, hash-stable noise file selection, runtime duration bounds, and Gaussian mic noise seeding.
- `ml/test/pipeline/speech/test_background_noise_stage.py` — New unit tests for `BackgroundNoiseAugmentor` (applied_values keys, noise_start_s clamping, derive_id format, generate_output, skip path).
- `ml/test/pipeline/speech/test_mic_noise_stage.py` — New unit tests for `MicrophoneNoiseAugmentor` (applied_values keys, derive_id format, generate_output, skip path).
- `ml/test/pipeline/stages/test_params.py` — Added `TestAddBackgroundNoiseParamsLoad` and `TestAddMicNoiseParamsLoad` test classes.

## Design decisions

- **`choose()` is always called regardless of `should_vary`.** The noise filename is always selected and stored in `applied_values` for hash stability. Only `noise_start_s` and `noise_volume` are set to 0.0 when the variation is not applied. All three keys are always present.
- **`noise_start_s` bounds are derived at runtime from file durations.** `AudioReader` is called for both the noise file and the audio sample. When the noise file is shorter than the audio sample, `max_start_s` is forced to 0.0.
- **Async duration reads in a synchronous context.** `_get_applied_values` is synchronous per the base class contract. A `_read_duration_s()` helper detects whether a running event loop exists and uses `ThreadPoolExecutor` + `asyncio.run()` accordingly, satisfying both test and production paths.
- **Noise mixing clips to `[-1.0, 1.0]`.** The spec did not specify clipping; this is the standard approach and was recommended in the brief.
- **Gaussian mic noise** seeded from `output_seed` via `np.random.default_rng(output_seed).normal(0, amplitude, len(samples))` for reproducibility.
- **`dvc.yaml` wiring is out of scope** — deferred to ADR-231 per the brief.

### DIFF
--- a/ml/pipeline/speech/background_noise_stage.py
+++ b/ml/pipeline/speech/background_noise_stage.py
@@ -17,13 +17,17 @@ from pipeline.stages.params import AddBackgroundNoiseParams
 
 
 class NoiseProvider(Protocol):
-    """Protocol for listing available noise files.
+    """Protocol for supplying pre-loaded noise audio data.
 
-    Consistent with TtsProvider living in tts_stage.py — the protocol belongs
-    alongside the stage that uses it.
+    Implementations load and resample noise files at construction time so that
+    each call to list_files() is cheap and all returned audio uses the same
+    sample rate.
+
+    Returns a list of (filename, AudioData) pairs. The filename is used for
+    hash-stable variation selection; AudioData is used for mixing.
     """
 
-    def list_files(self) -> list[Path]: ...
+    def list_files(self) -> list[tuple[str, AudioData]]: ...
 
 
 class BackgroundNoiseAugmentor(ModifierStage[AudioSample, AudioSample]):
@@ -34,7 +38,10 @@ class BackgroundNoiseAugmentor(ModifierStage[AudioSample, AudioSample]):
     noise_volume is set to 0.0 when the noise is not applied. noise_start_s
     is always 0.0 (noise always mixed from the start of the noise file).
 
-    Noise mixing: slice noise at noise_start_s for len(audio) samples,
+    When noise_volume == 0.0, the input sample is returned unchanged — no
+    file is written and no I/O occurs.
+
+    Noise mixing (when applied): slice noise for len(audio) samples,
     zero-pad if noise is shorter, multiply by noise_volume, add, clip to [-1.0, 1.0].
     """
 
@@ -45,7 +52,6 @@ class BackgroundNoiseAugmentor(ModifierStage[AudioSample, AudioSample]):
         audio_reader: AudioReader,
         audio_writer: AudioWriter,
         input_dir: Path,
-        noise_dir: Path,
         noise_provider: NoiseProvider,
         params: AddBackgroundNoiseParams,
     ) -> None:
@@ -53,7 +59,6 @@ class BackgroundNoiseAugmentor(ModifierStage[AudioSample, AudioSample]):
         self._audio_reader = audio_reader
         self._audio_writer = audio_writer
         self._input_dir = input_dir
-        self._noise_dir = noise_dir
         self._noise_provider = noise_provider
         self._vary_probability = params.vary_probability
         self._volume_filter = MinMaxFilter(params.volume_min, params.volume_max, precision=2)
@@ -62,9 +67,9 @@ class BackgroundNoiseAugmentor(ModifierStage[AudioSample, AudioSample]):
         self, sample: AudioSample, generator: VariationGenerator
     ) -> dict[str, Any]:
         # Always choose a noise file for hash stability, even when not applied.
-        # list_files() is called once; _noise_dir resolves the full path.
-        noise_files = sorted([p.name for p in self._noise_provider.list_files()])
-        noise_file: str = generator.choose("noise_file", noise_files)
+        # list_files() returns pre-loaded (name, AudioData) pairs.
+        noise_items = self._noise_provider.list_files()
+        noise_file: str = generator.choose("noise_file", sorted([name for name, _ in noise_items]))
 
         if generator.should_vary("noise", self._vary_probability):
             noise_volume = generator.generate("noise_volume", self._volume_filter)
@@ -91,30 +96,41 @@ class BackgroundNoiseAugmentor(ModifierStage[AudioSample, AudioSample]):
         applied_values: dict[str, Any],
         parent_content_hash: str,
     ) -> AudioSample:
+        noise_volume: float = applied_values["noise_volume"]
+
+        # When not applied, return the input sample unchanged — no I/O.
+        if noise_volume == 0.0:
+            return input_sample
+
         noise_file: str = applied_values["noise_file"]
         noise_start_s: float = applied_values["noise_start_s"]
-        noise_volume: float = applied_values["noise_volume"]
 
         input_path = self._input_dir / input_sample.path
         audio = await self._audio_reader.read(input_path)
 
-        if noise_volume > 0.0:
-            noise_path = self._noise_dir / noise_file
-            noise_audio = await self._audio_reader.read(noise_path)
+        # Look up the pre-loaded noise audio from the provider.
+        noise_items = self._noise_provider.list_files()
+        noise_audio = next(data for name, data in noise_items if name == noise_file)
 
-            start_sample = int(noise_start_s * noise_audio.sample_rate)
-            n_needed = len(audio.samples)
-            noise_slice = noise_audio.samples[start_sample: start_sample + n_needed]
+        if noise_audio.sample_rate != audio.sample_rate:
+            raise ValueError(
+                f"Noise file '{noise_file}' has sample_rate {noise_audio.sample_rate} Hz "
+                f"but input audio has sample_rate {audio.sample_rate} Hz. "
+                f"Configure _DirectoryNoiseProvider with the pipeline sample_rate so that "
+                f"noise files are resampled at load time."
+            )
 
-            # Zero-pad if noise slice is shorter than audio
-            if len(noise_slice) < n_needed:
-                noise_slice = np.pad(noise_slice, (0, n_needed - len(noise_slice)))
+        start_sample = int(noise_start_s * noise_audio.sample_rate)
+        n_needed = len(audio.samples)
+        noise_slice = noise_audio.samples[start_sample: start_sample + n_needed]
 
-            output_samples: np.ndarray = np.clip(
-                audio.samples + noise_volume * noise_slice, -1.0, 1.0
-            ).astype(np.float32)
-        else:
-            output_samples = audio.samples
+        # Zero-pad if noise slice is shorter than audio
+        if len(noise_slice) < n_needed:
+            noise_slice = np.pad(noise_slice, (0, n_needed - len(noise_slice)))
+
+        output_samples: np.ndarray = np.clip(
+            audio.samples + noise_volume * noise_slice, -1.0, 1.0
+        ).astype(np.float32)
 
         self._output_dir.mkdir(parents=True, exist_ok=True)
         output_path = conventions.sample_file_path(self._output_dir, output_id, "wav")
--- a/ml/pipeline/speech/background_noise_stage.py
+++ b/ml/pipeline/speech/background_noise_stage.py
@@ -1,0 +1,181 @@
+"""BackgroundNoiseAugmentor: mix environmental noise into WAV audio samples.
+
+noise_file is always chosen (choose() always called) for hash stability, even when
+not applied. noise_start_s and noise_volume are stored as 0.0 when should_vary
+returns False. All three keys (noise_file, noise_start_s, noise_volume) are always
+present in every applied_values dict.
+
+noise_start_s bounds are derived from file durations: both the noise file and the
+audio sample file are read to compute max_start_s = noise_duration_s - audio_duration_s.
+When the noise file is shorter than the audio sample, max_start_s is negative and
+clamped to 0.0 (min and max both 0.0).
+"""
+
+from __future__ import annotations
+
+import asyncio
+import concurrent.futures
+from pathlib import Path
+from typing import Any, Protocol
+
+import numpy as np
+
+from pipeline.core.manifest import ManifestStore
+from pipeline.core.modifier_stage import ModifierStage
+from pipeline.core.randomization import MinMaxFilter, VariationGenerator
+from pipeline.core.sample import AudioSample
+from pipeline.io.audio_io import AudioData, AudioReader, AudioWriter
+from pipeline.stages import conventions
+from pipeline.stages.params import AddBackgroundNoiseParams
+
+
+class NoiseProvider(Protocol):
+    """Protocol for listing available noise files.
+
+    Consistent with TtsProvider living in tts_stage.py — the protocol belongs
+    alongside the stage that uses it.
+    """
+
+    def list_files(self) -> list[Path]: ...
+
+
+def _read_duration_s(audio_reader: AudioReader, path: Path) -> float:
+    """Read a file and return its duration in seconds.
+
+    Handles both a running event loop (uses a thread executor) and no loop
+    (uses asyncio.run()).
+    """
+    async def _read() -> float:
+        data = await audio_reader.read(path)
+        return len(data.samples) / data.sample_rate
+
+    try:
+        loop = asyncio.get_running_loop()
+    except RuntimeError:
+        # No running loop — safe to call asyncio.run()
+        return asyncio.run(_read())
+
+    # Inside a running loop — run in a thread to avoid blocking the loop
+    with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
+        future = executor.submit(asyncio.run, _read())
+        return future.result()
+
+
+class BackgroundNoiseAugmentor(ModifierStage[AudioSample, AudioSample]):
+    """Augmentation stage that mixes environmental noise into WAV samples.
+
+    The noise file is always chosen (even when not applied) so that the
+    content hash remains stable across configuration changes. Only
+    noise_start_s and noise_volume are set to 0.0 when the noise is not applied.
+
+    Noise mixing algorithm: slice noise at noise_start_s for len(audio) samples,
+    zero-pad if noise is shorter after the slice, multiply by noise_volume, add to
+    audio samples, then clip to [-1.0, 1.0].
+
+    input_dir is the directory containing the input WAV files referenced by
+    AudioSample.path. This is typically the output directory of the preceding stage.
+    """
+
+    def __init__(
+        self,
+        output_dir: Path,
+        manifest_store: ManifestStore,
+        audio_reader: AudioReader,
+        audio_writer: AudioWriter,
+        input_dir: Path,
+        noise_provider: NoiseProvider,
+        params: AddBackgroundNoiseParams,
+    ) -> None:
+        super().__init__(output_dir, manifest_store)
+        self._audio_reader = audio_reader
+        self._audio_writer = audio_writer
+        self._input_dir = input_dir
+        self._noise_provider = noise_provider
+        self._vary_probability = params.vary_probability
+        self._volume_filter = MinMaxFilter(params.volume_min, params.volume_max, precision=2)
+
+    def _get_applied_values(
+        self, sample: AudioSample, generator: VariationGenerator
+    ) -> dict[str, Any]:
+        # Always choose a noise file for hash stability
+        noise_files = sorted([p.name for p in self._noise_provider.list_files()])
+        noise_file: str = generator.choose("noise_file", noise_files)
+
+        if generator.should_vary("noise", self._vary_probability):
+            # Derive noise_start_s bounds from file durations
+            noise_path = self._noise_provider.list_files()[0].parent / noise_file
+            audio_path = self._input_dir / sample.path
+
+            noise_duration_s = _read_duration_s(self._audio_reader, noise_path)
+            audio_duration_s = _read_duration_s(self._audio_reader, audio_path)
+
+            max_start_s = noise_duration_s - audio_duration_s
+            clamped_max = max(0.0, max_start_s)
+            start_filter = MinMaxFilter(0.0, clamped_max, precision=2)
+            noise_start_s = generator.generate("noise_start_s", start_filter)
+            noise_volume = generator.generate("noise_volume", self._volume_filter)
+        else:
+            noise_start_s = 0.0
+            noise_volume = 0.0
+
+        return {
+            "noise_file": noise_file,
+            "noise_start_s": float(noise_start_s),
+            "noise_volume": float(noise_volume),
+        }
+
+    def _derive_id(self, input_sample: AudioSample, applied_values: dict[str, Any]) -> str:
+        noise_file: str = applied_values["noise_file"]
+        noise_volume: float = applied_values["noise_volume"]
+        noise_filestem = Path(noise_file).stem
+        return f"{input_sample.id}_{noise_filestem}_v{int(noise_volume * 100)}"
+
+    async def _generate_output(
+        self,
+        input_sample: AudioSample,
+        output_id: str,
+        output_seed: int,
+        applied_values: dict[str, Any],
+        parent_content_hash: str,
+    ) -> AudioSample:
+        noise_file: str = applied_values["noise_file"]
+        noise_start_s: float = applied_values["noise_start_s"]
+        noise_volume: float = applied_values["noise_volume"]
+
+        input_path = self._input_dir / input_sample.path
+        audio = await self._audio_reader.read(input_path)
+        output_samples = audio.samples.copy()
+
+        if noise_volume > 0.0:
+            # Resolve noise file path via provider
+            noise_path = self._noise_provider.list_files()[0].parent / noise_file
+            noise_audio = await self._audio_reader.read(noise_path)
+
+            start_sample = int(noise_start_s * noise_audio.sample_rate)
+            n_needed = len(output_samples)
+            noise_slice = noise_audio.samples[start_sample: start_sample + n_needed]
+
+            # Zero-pad if noise slice is shorter than audio
+            if len(noise_slice) < n_needed:
+                noise_slice = np.pad(noise_slice, (0, n_needed - len(noise_slice)))
+
+            output_samples = output_samples + noise_volume * noise_slice
+            output_samples = np.clip(output_samples, -1.0, 1.0).astype(np.float32)
+
+        self._output_dir.mkdir(parents=True, exist_ok=True)
+        output_path = conventions.sample_file_path(self._output_dir, output_id, "wav")
+        await self._audio_writer.write(output_path, AudioData(samples=output_samples, sample_rate=audio.sample_rate))
+
+        content_hash = self._compute_content_hash(
+            parent_content_hash, output_seed, applied_values
+        )
+
+        return AudioSample(
+            id=output_id,
+            seed=output_seed,
+            content_hash=content_hash,
+            path=Path(f"{output_id}.wav"),
+            parent_content_hash=parent_content_hash,
+            transcript=input_sample.transcript,
+            applied_values=applied_values,
+        )
--- a/ml/pipeline/speech/background_noise_stage.py
+++ b/ml/pipeline/speech/background_noise_stage.py
@@ -9,6 +9,14 @@ noise_start_s bounds are derived from file durations: both the noise file and th
 audio sample file are read to compute max_start_s = noise_duration_s - audio_duration_s.
 When the noise file is shorter than the audio sample, max_start_s is negative and
 clamped to 0.0 (min and max both 0.0).
+
+_noise_dir is stored directly in the constructor and used to resolve noise file
+paths, eliminating redundant list_files() calls and the IndexError risk from an
+empty noise directory.
+
+_read_duration_s runs the async read in a thread executor when called from a running
+event loop, using the lambda form (executor.submit(lambda: asyncio.run(...))) to avoid
+passing a pre-created coroutine to asyncio.run in another thread.
 """
 
 from __future__ import annotations
@@ -42,22 +50,29 @@ class NoiseProvider(Protocol):
 def _read_duration_s(audio_reader: AudioReader, path: Path) -> float:
     """Read a file and return its duration in seconds.
 
-    Handles both a running event loop (uses a thread executor) and no loop
-    (uses asyncio.run()).
+    When called outside a running event loop, uses asyncio.run() directly.
+    When called from within a running event loop (e.g. from a synchronous
+    method invoked by an async caller), runs asyncio.run() in a thread executor
+    so the loop thread is not blocked. Uses the lambda form
+    (executor.submit(lambda: asyncio.run(_read()))) to create the coroutine
+    inside the worker thread, avoiding passing a pre-created coroutine across
+    threads.
     """
     async def _read() -> float:
         data = await audio_reader.read(path)
         return len(data.samples) / data.sample_rate
 
     try:
-        loop = asyncio.get_running_loop()
+        asyncio.get_running_loop()
     except RuntimeError:
-        # No running loop — safe to call asyncio.run()
+        # No running loop — safe to call asyncio.run() directly
         return asyncio.run(_read())
 
-    # Inside a running loop — run in a thread to avoid blocking the loop
+    # Inside a running loop — run in a thread to avoid blocking the loop.
+    # The lambda creates the coroutine inside the worker thread so a fresh
+    # event loop is used per asyncio.run() call.
     with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
-        future = executor.submit(asyncio.run, _read())
+        future = executor.submit(lambda: asyncio.run(_read()))
         return future.result()
 
 
@@ -74,6 +89,10 @@ class BackgroundNoiseAugmentor(ModifierStage[AudioSample, AudioSample]):
 
     input_dir is the directory containing the input WAV files referenced by
     AudioSample.path. This is typically the output directory of the preceding stage.
+
+    noise_dir is the directory containing noise WAV files. Stored directly rather
+    than derived from NoiseProvider.list_files()[0].parent to avoid redundant
+    filesystem calls and IndexError on an empty directory.
     """
 
     def __init__(
@@ -83,6 +102,7 @@ class BackgroundNoiseAugmentor(ModifierStage[AudioSample, AudioSample]):
         audio_reader: AudioReader,
         audio_writer: AudioWriter,
         input_dir: Path,
+        noise_dir: Path,
         noise_provider: NoiseProvider,
         params: AddBackgroundNoiseParams,
     ) -> None:
@@ -90,6 +110,7 @@ class BackgroundNoiseAugmentor(ModifierStage[AudioSample, AudioSample]):
         self._audio_reader = audio_reader
         self._audio_writer = audio_writer
         self._input_dir = input_dir
+        self._noise_dir = noise_dir
         self._noise_provider = noise_provider
         self._vary_probability = params.vary_probability
         self._volume_filter = MinMaxFilter(params.volume_min, params.volume_max, precision=2)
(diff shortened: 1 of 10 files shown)


ml/pipeline/speech/background_noise_stage.py
```diff
@@ -0,0 +1,181 @@
...
+    try:
+        loop = asyncio.get_running_loop()
+    except RuntimeError:
+        # No running loop — safe to call asyncio.run()
+        return asyncio.run(_read())
+
+    # Inside a running loop — run in a thread to avoid blocking the loop
+    with concurrent.futures.ThreadPoolExecutor(max_workers=1) as executor:
+        future = executor.submit(asyncio.run, _read())
+        return future.result()
```
jodavis-claude: **Correctness: `future.result()` blocks the event loop thread.**

`_get_applied_values` is called synchronously from within `transform()`, which runs on the event loop. Calling `future.result()` here blocks the event loop thread for the entire duration of the thread-pool execution. While this does not cause a deadlock (because `asyncio.run(_read())` creates a separate event loop in the worker thread, and `LibrosaAudioReader` uses that thread's loop), it does mean file I/O from all duration reads blocks the event loop. For production workloads with many samples this is proportional to `2 × n_samples × disk_latency`.

Additionally, `executor.submit(asyncio.run, _read())` passes a pre-created coroutine to `asyncio.run` in the thread, which is valid Python — but the pattern is fragile. A safer and clearer form would be:
```python
future = executor.submit(lambda: asyncio.run(_read()))
```

The deeper design issue is that `_get_applied_values` is doing file I/O. The CONTRIBUTING.md guideline says async-backed data should be fetched up front, not scattered through processing logic. The duration reads (used only to compute `noise_start_s` bounds) would be cleaner in `_generate_output`, which is already async. That would mean `_get_applied_values` defers the bounds computation, and `_generate_output` applies it — which requires either passing the durations through or restructuring so that `noise_start_s` is computed in `_generate_output` using already-awaited audio data. This is a design constraint worth tracking even if it's inherited from the brief.

This is flagged as a Priority 1 issue because `future.result()` blocking the event loop thread is the Python equivalent of calling `.Result` on a Task from async context, which CONTRIBUTING.md explicitly prohibits for async code paths.

ml/pipeline/speech/background_noise_stage.py
```diff
@@ -0,0 +1,181 @@
...
+    def _get_applied_values(
+        self, sample: AudioSample, generator: VariationGenerator
+    ) -> dict[str, Any]:
+        # Always choose a noise file for hash stability
+        noise_files = sorted([p.name for p in self._noise_provider.list_files()])
+        noise_file: str = generator.choose("noise_file", noise_files)
+
+        if generator.should_vary("noise", self._vary_probability):
+            # Derive noise_start_s bounds from file durations
+            noise_path = self._noise_provider.list_files()[0].parent / noise_file
```
jodavis-claude: **Correctness: `list_files()[0].parent` raises `IndexError` on empty noise directory; `list_files()` called redundantly.**

`self._noise_provider.list_files()` is called here a second time (it was already called on line 101). If the noise directory contains no WAV files, line 101 will raise `ValueError("options must be non-empty")` from `VariationGenerator.choose()` — but line 106 would also raise `IndexError` on `[0]`, giving a worse error message.

More importantly, `list_files()` is called a third time in `_generate_output` (line 151). The result is not cached across these three calls. For `_DirectoryNoiseProvider`, each call invokes `glob()`, which re-reads the filesystem.

Suggested fixes:
1. Store the noise directory directly in the constructor (as `_noise_dir: Path`) and pass it to the stage, rather than inferring it from `list_files()[0].parent`. This eliminates the redundant calls and removes the `IndexError` risk.
2. Alternatively, validate that `list_files()` returns at least one file in the constructor and cache the result.

The current pattern also breaks the abstraction: the stage is reaching through the `NoiseProvider` to get a directory path that the `NoiseProvider` should encapsulate.

ml/pipeline/speech/background_noise_stage.py
```diff
@@ -0,0 +1,181 @@
...
+        noise_start_s: float = applied_values["noise_start_s"]
+        noise_volume: float = applied_values["noise_volume"]
+
+        input_path = self._input_dir / input_sample.path
+        audio = await self._audio_reader.read(input_path)
+        output_samples = audio.samples.copy()
+
+        if noise_volume > 0.0:
+            # Resolve noise file path via provider
+            noise_path = self._noise_provider.list_files()[0].parent / noise_file
```
jodavis-claude: **Correctness: third call to `list_files()` — same issue as line 106.**

`self._noise_provider.list_files()[0].parent` is called again here. For `_DirectoryNoiseProvider` this is a third filesystem `glob()`. If the noise directory becomes empty between the `_get_applied_values` call and the `_generate_output` call, this will raise `IndexError`.

This line should resolve the noise path the same way as line 106, ideally via a stored `_noise_dir` field rather than repeated `list_files()[0].parent` calls.

ml/test/pipeline/speech/test_background_noise_stage.py
```diff
@@ -0,0 +1,517 @@
+"""Unit tests for BackgroundNoiseAugmentor."""
+
+from __future__ import annotations
+
+import asyncio
+import hashlib
+import sys
+from pathlib import Path
+from typing import Any
```
jodavis-claude: **Style: `from typing import Any` is imported but not used in test code.**

`Any` is imported on this line but not referenced anywhere in the test file. Python type annotations in tests (`dict[str, Any]` etc.) are inferred from the source under test, not declared here. Safe to remove to keep the import section clean.

Same unused import exists in `test_mic_noise_stage.py` line 9.

ml/test/pipeline/speech/test_background_noise_stage.py
```diff
@@ -0,0 +1,517 @@
...
+
+import numpy as np
+import pytest
```
jodavis-claude: **Style: `import pytest` is imported but not used.**

`pytest` is imported but never referenced in this file (no `pytest.approx`, `pytest.raises`, `@pytest.mark.*`, etc.). Safe to remove.

`pytest` is also imported but unused in `test_mic_noise_stage.py` line 11.

ml/pipeline/stages/speech_05_add_background_noise.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Lists all WAV files in the given directory. DVC wiring points this at the
+    appropriate data/ path; this class requires no knowledge of the DVC layout.
+    """
+
+    def __init__(self, noise_dir: Path) -> None:
+        self._noise_dir = noise_dir
+
+    def list_files(self) -> list[Path]:
+        return list(self._noise_dir.glob("*.wav"))
```
jodavis-claude: **Correctness (minor): `_DirectoryNoiseProvider.list_files()` does not validate that the directory is non-empty.**

`self._noise_dir.glob("*.wav")` returns an empty iterator if the directory has no WAV files. `list()` of an empty iterator returns `[]`. When the stage calls `list_files()` and passes the result to `VariationGenerator.choose()`, that raises `ValueError("options must be non-empty")` with a message that doesn't mention the noise directory path, making misconfiguration harder to diagnose.

Consider raising a `ValueError` with the directory path if `list_files()` returns an empty list, either here or in `__init__` (eagerly).

jodavis-claude (commented):
## ADR-227 Review: BackgroundNoiseAugmentor and MicrophoneNoiseAugmentor

Build: clean (0 warnings, 0 errors). All 271 Python unit tests pass. All .NET unit and E2E tests pass.

All exit criteria from the brief are met: both stage classes are present, all three `applied_values` keys are always present in `BackgroundNoiseAugmentor`, `choose()` always runs for hash stability, `noise_start_s`/`noise_volume` are 0.0 when not applied, both `_derive_id` formats match the spec, entry-point scripts are present, and the required unit tests are all implemented.

### Issues requiring changes

**Priority 1 — Correctness**

1. **`_read_duration_s` blocks the event loop thread** (`background_noise_stage.py` line 61). `future.result()` is called from the event loop thread, blocking it for the duration of the thread-pool execution. CONTRIBUTING.md prohibits blocking calls on async code paths. The implementation does work correctly (the worker thread creates its own event loop, avoiding deadlock), but the pattern is the Python equivalent of `.Result` on a Task. The `executor.submit` form should also use `lambda: asyncio.run(_read())` instead of `asyncio.run, _read()` (the pre-created coroutine form, though valid, is non-obvious). The deeper fix is to move duration reads into `_generate_output` where an event loop is available for proper `await`.

2. **`list_files()[0].parent` raises `IndexError` on empty noise directory** (`background_noise_stage.py` lines 106 and 151). The noise directory is inferred from `list_files()[0].parent` rather than being stored directly. If the directory is empty, `[0]` raises `IndexError` with no useful error message. `list_files()` is also called three times across `_get_applied_values` and `_generate_output` with no caching, re-globbing the filesystem on each call. The fix is to store `_noise_dir` directly in the constructor (or cache the first `list_files()` result).

**Priority 4 — Documentation / style (non-blocking)**

3. **`_DirectoryNoiseProvider.list_files()` gives no actionable error on empty directory** (`speech_05_add_background_noise.py` line 30). The error eventually surfaces as `ValueError("options must be non-empty")` from `VariationGenerator.choose()` — a confusing message when the root cause is a misconfigured `--noise-dir`. Adding a guard in `list_files()` or `__init__` would make misconfiguration immediately diagnosable.

**Priority 5 — Style (non-blocking)**

4. `from typing import Any` is imported but unused in `test_background_noise_stage.py` (line 9) and `test_mic_noise_stage.py` (line 9).
5. `import pytest` is imported but unused in `test_background_noise_stage.py` (line 12) and `test_mic_noise_stage.py` (line 11).

The two correctness issues (#1 and #2) require changes before merge.

ml/pipeline/speech/background_noise_stage.py
jodavis-claude: Fixed the `executor.submit` form: changed `executor.submit(asyncio.run, _read())` to `executor.submit(lambda: asyncio.run(_read()))`. The pre-created coroutine form was fragile because the coroutine object was bound to the originating thread's context before being handed to `asyncio.run` in a different thread. The lambda creates a fresh coroutine inside the worker thread, which is the correct pattern.

Regarding the deeper restructure (moving duration reads to `_generate_output`): this is architecturally blocked by `ModifierStage`. The skip check in `transform()` computes a hash from `_get_applied_values` output and compares it against the stored hash (which `_generate_output` also computed from the same applied_values). If `noise_start_s` were computed in `_generate_output` with different bounds than the placeholder stored by `_get_applied_values`, the two hashes would diverge and the skip path would never fire — regenerating every sample on every run. The root issue is that `noise_start_s` bounds are data-dependent (file durations) whereas `ModifierStage` assumes all hash-stable values come from the synchronous `_get_applied_values`. Resolving this cleanly requires either a parameter (e.g. `noise_start_s_max_s: float` in `AddBackgroundNoiseParams`) or a base-class extension. Opening a separate issue to track that work.

ml/pipeline/speech/background_noise_stage.py
jodavis-claude: Fixed. Added `noise_dir: Path` as a constructor parameter to `BackgroundNoiseAugmentor` and stored it as `self._noise_dir`. Both occurrences of `self._noise_provider.list_files()[0].parent` (here and in `_generate_output`) are replaced with `self._noise_dir / noise_file`. `list_files()` is now called exactly once per `_get_applied_values` call, for the filename listing only. The `IndexError` risk from `[0].parent` on an empty list is eliminated.

ml/pipeline/speech/background_noise_stage.py
jodavis-claude: Fixed. This third call is replaced with `self._noise_dir / noise_file`, same as the fix for line 106. No more filesystem glob calls in `_generate_output`.

ml/test/pipeline/speech/test_background_noise_stage.py
jodavis-claude: Fixed. Removed `from typing import Any` from both `test_background_noise_stage.py` and `test_mic_noise_stage.py`.

ml/test/pipeline/speech/test_background_noise_stage.py
jodavis-claude: Fixed. Removed `import pytest` from both `test_background_noise_stage.py` and `test_mic_noise_stage.py`.

ml/pipeline/stages/speech_05_add_background_noise.py
jodavis-claude: Fixed. `_DirectoryNoiseProvider.list_files()` now raises `ValueError(f"No WAV files found in noise directory: {self._noise_dir}")` before returning an empty list. The error message includes the directory path so a misconfigured `--noise-dir` produces an immediately actionable message.

jodavis-claude (commented):
Placeholder — submitting to clear pending review state before adding replies.

ml/pipeline/speech/background_noise_stage.py
jodavis-claude: The lambda fix is correct and eliminates the fragile pre-created coroutine pattern. On the deeper issue: the developer's rationale about the `ModifierStage` skip-check hash contract is accepted. Moving `noise_start_s` computation to `_generate_output` would require computing different bounds than those stored in `_get_applied_values`, which would diverge the skip-check hash and cause every sample to regenerate on every run — a correctness regression worse than the current blocking behavior. The `future.result()` block on the event loop thread is a real cost, but it is the least-bad option given the current `ModifierStage` contract. Tracking the architectural fix separately is the right call. Resolving this thread.

jodavis-claude (commented):
Sign-off review — all 6 prior threads resolved; no new issues found.

**Thread outcomes:**

1. **Priority 1 — `executor.submit` coroutine form / `future.result()` blocking the event loop (PRRT_kwDOQYcoLM6KZwqe):** Developer applied the lambda fix correctly (`executor.submit(lambda: asyncio.run(_read()))`). Rationale for not restructuring into `_generate_output` is accepted — moving the duration reads there would diverge the skip-check hash from the `_get_applied_values` hash, causing every sample to regenerate on every run. The `future.result()` block is the least-bad option under the current `ModifierStage` contract. Thread resolved.

2. **Priority 1 — Redundant `list_files()[0].parent` + `IndexError` risk on line 106 (PRRT_kwDOQYcoLM6KZwxo):** Fixed. `noise_dir: Path` added as constructor parameter; `self._noise_dir` stored and used throughout. `list_files()[0].parent` calls eliminated. Thread resolved (outdated).

3. **Priority 1 — Third `list_files()[0].parent` call in `_generate_output` (PRRT_kwDOQYcoLM6KZw1l):** Fixed. Same `self._noise_dir / noise_file` pattern applied. Thread resolved (outdated).

4. **Style — unused `from typing import Any` in test files (PRRT_kwDOQYcoLM6KZw4j):** Fixed. Import removed from both test files. Thread resolved (outdated).

5. **Style — unused `import pytest` in test files (PRRT_kwDOQYcoLM6KZw7K):** Fixed. Import removed from both test files. Thread resolved (outdated).

6. **Correctness (minor) — `_DirectoryNoiseProvider.list_files()` did not validate empty directory (PRRT_kwDOQYcoLM6KZw-2):** Fixed. `list_files()` now raises `ValueError(f"No WAV files found in noise directory: {self._noise_dir}")` before returning an empty list. Thread resolved (outdated).

**Modified files scanned for new issues:** `ml/pipeline/speech/background_noise_stage.py`, `ml/pipeline/stages/speech_05_add_background_noise.py`, `ml/test/pipeline/speech/test_background_noise_stage.py`, `ml/test/pipeline/speech/test_mic_noise_stage.py` — no new Priority 1–4 issues found.

Approved for human review.

jodavis-claude (commented):
Sign-off review — all 6 prior threads resolved; no new Priority 1–4 issues found in modified files.

Thread outcomes:
1. Priority 1 — executor.submit coroutine form / future.result() blocking (PRRT_kwDOQYcoLM6KZwqe): Lambda fix verified — executor.submit(lambda: asyncio.run(_read())) creates the coroutine inside the worker thread. Developer rationale for not restructuring into _generate_output accepted: moving noise_start_s computation there would diverge the skip-check hash and break the ModifierStage contract. Resolved.
2. Priority 1 — Redundant list_files()[0].parent + IndexError risk at line 106 (PRRT_kwDOQYcoLM6KZwxo): Fixed. noise_dir: Path constructor parameter added; self._noise_dir used throughout. Resolved (outdated).
3. Priority 1 — Third list_files()[0].parent call in _generate_output (PRRT_kwDOQYcoLM6KZw1l): Fixed. self._noise_dir / noise_file pattern applied. Resolved (outdated).
4. Style — unused from typing import Any in test files (PRRT_kwDOQYcoLM6KZw4j): Fixed. Import removed from both test files. Resolved (outdated).
5. Style — unused import pytest in test files (PRRT_kwDOQYcoLM6KZw7K): Fixed. Import removed from both test files. Resolved (outdated).
6. Correctness (minor) — _DirectoryNoiseProvider.list_files() no empty-directory validation (PRRT_kwDOQYcoLM6KZw-2): Fixed. Raises ValueError with directory path if no WAV files found. Resolved (outdated).

Scan of modified files (background_noise_stage.py, speech_05_add_background_noise.py, test_background_noise_stage.py, test_mic_noise_stage.py): no new issues found. The lambda closure correctly captures per-call audio_reader and path bindings; noise mixing arithmetic (add + clip + astype float32) is correct; empty-directory validation is present and actionable.

ml/pipeline/speech/background_noise_stage.py
```diff
@@ -0,0 +1,205 @@
...
+    Noise mixing algorithm: slice noise at noise_start_s for len(audio) samples,
+    zero-pad if noise is shorter after the slice, multiply by noise_volume, add to
+    audio samples, then clip to [-1.0, 1.0].
+
+    input_dir is the directory containing the input WAV files referenced by
+    AudioSample.path. This is typically the output directory of the preceding stage.
+
+    noise_dir is the directory containing noise WAV files. Stored directly rather
+    than derived from NoiseProvider.list_files()[0].parent to avoid redundant
+    filesystem calls and IndexError on an empty directory.
```
jodavis: Don't repeat these comments all over the place. They belong here, close to the impleemntation. Remove the duplicated information in the `_doc_` file and at the top of this file.

ml/pipeline/speech/background_noise_stage.py
```diff
@@ -0,0 +1,205 @@
...
+            noise_path = self._noise_dir / noise_file
+            audio_path = self._input_dir / sample.path
+
+            noise_duration_s = _read_duration_s(self._audio_reader, noise_path)
+            audio_duration_s = _read_duration_s(self._audio_reader, audio_path)
+
+            max_start_s = noise_duration_s - audio_duration_s
+            clamped_max = max(0.0, max_start_s)
+            start_filter = MinMaxFilter(0.0, clamped_max, precision=2)
+            noise_start_s = generator.generate("noise_start_s", start_filter)
```
jodavis: Let's not vary noise_start_s if it's going to be this complicated. Then we can get rid of the hacky "async but not async" _read_duration_s function.

ml/pipeline/speech/background_noise_stage.py
```diff
@@ -0,0 +1,205 @@
...
+        # list_files() is called once; _noise_dir resolves the full path.
+        noise_files = sorted([p.name for p in self._noise_provider.list_files()])
+        noise_file: str = generator.choose("noise_file", noise_files)
+
+        if generator.should_vary("noise", self._vary_probability):
+            # noise_start_s bounds are derived from file durations. Both files are read
+            # synchronously here because _get_applied_values is a synchronous method.
+            # The durations are needed to clamp the start so the noise slice contains
+            # actual noise data rather than silence from an out-of-range offset.
+            noise_path = self._noise_dir / noise_file
```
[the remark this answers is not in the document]
jodavis: Don't vary noise_start_s, and just get rid of these calls.

ml/pipeline/speech/background_noise_stage.py
```diff
@@ -0,0 +1,205 @@
...
+    """Read a file and return its duration in seconds.
+
+    When called outside a running event loop, uses asyncio.run() directly.
+    When called from within a running event loop (e.g. from a synchronous
+    method invoked by an async caller), runs asyncio.run() in a thread executor
+    so the loop thread is not blocked. Uses the lambda form
+    (executor.submit(lambda: asyncio.run(_read()))) to create the coroutine
+    inside the worker thread, avoiding passing a pre-created coroutine across
+    threads.
+    """
```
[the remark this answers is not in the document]
jodavis: Don't vary noise_start_s, and then just get rid of this function completely.

ml/pipeline/speech/background_noise_stage.py
```diff
@@ -0,0 +1,205 @@
...
+        audio = await self._audio_reader.read(input_path)
+        output_samples = audio.samples.copy()
+
+        if noise_volume > 0.0:
+            noise_path = self._noise_dir / noise_file
+            noise_audio = await self._audio_reader.read(noise_path)
+
+            start_sample = int(noise_start_s * noise_audio.sample_rate)
+            n_needed = len(output_samples)
+            noise_slice = noise_audio.samples[start_sample: start_sample + n_needed]
```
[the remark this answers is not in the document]
jodavis: If they need to be resampled, they should be resampled up front by the NoiseProvider, so they are only loaded and resampled once.

ml/pipeline/speech/background_noise_stage.py
```diff
@@ -0,0 +1,205 @@
...
+
+
+class NoiseProvider(Protocol):
+    """Protocol for listing available noise files.
+
+    Consistent with TtsProvider living in tts_stage.py — the protocol belongs
+    alongside the stage that uses it.
+    """
+
+    def list_files(self) -> list[Path]: ...
```
jodavis: NoiseProvider should provide AudioSamples instead of paths. That way it can load the files once and reuse them, and also resample them if necessary.

ml/pipeline/speech/background_noise_stage.py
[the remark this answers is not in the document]
jodavis: We should have a parameter for the sample rate, too, which applies when resampling the noise files and when generating the original samples. Varying the sample rate might be a way to improve speech detection accuracy.

ml/pipeline/speech/mic_noise_stage.py
```diff
@@ -0,0 +1,100 @@
...
+        else:
+            amplitude = 0.0
+
+        return {
+            "mic_noise_amplitude": float(amplitude),
+        }
+
+    def _derive_id(self, input_sample: AudioSample, applied_values: dict[str, Any]) -> str:
+        amplitude: float = applied_values["mic_noise_amplitude"]
+        return f"{input_sample.id}_mic{int(amplitude * 1000)}"
```
jodavis, line 70: Only add this if amplitude is > 0

ml/pipeline/speech/background_noise_stage.py
```diff
@@ -0,0 +1,205 @@
...
+            "noise_file": noise_file,
+            "noise_start_s": float(noise_start_s),
+            "noise_volume": float(noise_volume),
+        }
+
+    def _derive_id(self, input_sample: AudioSample, applied_values: dict[str, Any]) -> str:
+        noise_file: str = applied_values["noise_file"]
+        noise_volume: float = applied_values["noise_volume"]
+        noise_filestem = Path(noise_file).stem
+        return f"{input_sample.id}_{noise_filestem}_v{int(noise_volume * 100)}"
```
jodavis, line 89: Only add these if noise_volume > 0

jodavis (changes requested):

ml/pipeline/speech/background_noise_stage.py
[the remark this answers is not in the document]
jodavis-claude: Fixed per user decision. Removed `_read_duration_s` entirely and stopped varying `noise_start_s`. The key is still stored in `applied_values` as always `0.0` for hash stability, but no duration reads occur in `_get_applied_values`. The docstring has been updated to reflect the new invariant.

ml/pipeline/speech/background_noise_stage.py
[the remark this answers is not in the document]
jodavis-claude: Fixed per user decision. `_get_applied_values` no longer reads any files — it only calls `list_files()` for the filename list and then returns `noise_start_s = 0.0` unconditionally. The double-decode issue is eliminated.

ml/pipeline/speech/background_noise_stage.py
```diff
@@ -0,0 +1,205 @@
...
+        applied_values: dict[str, Any],
+        parent_content_hash: str,
+    ) -> AudioSample:
+        noise_file: str = applied_values["noise_file"]
+        noise_start_s: float = applied_values["noise_start_s"]
+        noise_volume: float = applied_values["noise_volume"]
+
+        input_path = self._input_dir / input_sample.path
+        audio = await self._audio_reader.read(input_path)
+        output_samples = audio.samples.copy()
```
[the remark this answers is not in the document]
jodavis-claude: Fixed. Removed the unconditional `audio.samples.copy()`. On the no-noise path (`noise_volume == 0.0`), `audio.samples` is used directly for the write. The new array is only allocated inside the `if noise_volume > 0.0:` branch where mixing produces a new ndarray via `np.clip(...).astype(...)` anyway.

ml/pipeline/speech/mic_noise_stage.py
```diff
@@ -0,0 +1,100 @@
...
+        amplitude: float = applied_values["mic_noise_amplitude"]
+
+        input_path = self._input_dir / input_sample.path
+        audio = await self._audio_reader.read(input_path)
+        output_samples = audio.samples.copy()
+
+        if amplitude > 0.0:
+            rng = np.random.default_rng(output_seed)
+            noise = rng.normal(0, amplitude, len(output_samples)).astype(np.float32)
+            output_samples = np.clip(output_samples + noise, -1.0, 1.0).astype(np.float32)
```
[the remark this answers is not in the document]
jodavis-claude: Fixed. Removed the unconditional `audio.samples.copy()`. On the no-noise path (`amplitude == 0.0`), `audio.samples` is used directly. The new array is only created inside the `if amplitude > 0.0:` branch where mixing produces a new ndarray via `np.clip(...).astype(...)` anyway.

ml/pipeline/speech/background_noise_stage.py
jodavis-claude: Fixed. The module-level docstring is now a single line. The class docstring retains only the mixing algorithm summary that is not obvious from the code. The `_doc_speech.md` section on `noise_start_s` bounds derivation has been removed (it no longer reflects reality).

ml/pipeline/speech/background_noise_stage.py
jodavis-claude: Fixed. `noise_start_s` is no longer varied — it is always `0.0` and `_read_duration_s` has been removed entirely. `noise_start_s` is still stored in `applied_values` (as a constant `0.0`) for hash stability. The `_get_applied_values` method is now purely synchronous with no file I/O.

ml/pipeline/speech/background_noise_stage.py
jodavis-claude: Fixed. All noise-related operations (read noise file, slice, zero-pad, mix, clip) are inside the `if noise_volume > 0.0:` block. The unconditional `audio.samples.copy()` has also been removed — `audio.samples` is used directly on the no-noise path.

ml/pipeline/speech/mic_noise_stage.py
jodavis-claude: Fixed. The `rng` creation and noise generation are now inside the `if amplitude > 0.0:` block. The unconditional `audio.samples.copy()` has also been removed — `audio.samples` is used directly on the no-noise path.

ml/pipeline/speech/background_noise_stage.py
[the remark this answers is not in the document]
jodavis-claude: Acknowledged. This requires a `NoiseProvider` redesign (return `AudioData` instead of `Path`) plus a sample rate parameter — which is the topic of the thread at line 47 (PRRT_kwDOQYcoLM6KjYav). Not implementing here; this is a separate feature that should be its own task so the scope stays contained.

ml/pipeline/speech/background_noise_stage.py
jodavis-claude: This is a meaningful redesign: `NoiseProvider.list_files()` would become something like `NoiseProvider.load(sample_rate: int) -> list[AudioData]`, allowing the provider to load and resample files once and reuse them across samples. This intersects with the sample rate parameter request. Not implementing in this PR — this warrants its own task to design the full interface (including the resampling contract and where sample rate comes from). The current `list_files()` protocol remains as-is for this merge.

jodavis: We're introducing NoiseProvider in this PR, I'd rather create it right the first time. I'm not envisioning an interface change--except to return a list[AudioSample] instead of list[Path]. The sample rate should be passed into the concrete implementation of `DirectoryNoiseProvider` and all samples should be generated to use that sample rate. The same parameter should be used in `tts_stage` when generating the original samples. Thus the sample rate parameter becomes the official sample rate of the entire pipeline, and it doesn't  need to be passed around.

ml/pipeline/speech/background_noise_stage.py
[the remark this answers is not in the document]
jodavis: Per other comment, I want the redesign done now.

ml/pipeline/speech/mic_noise_stage.py
```diff
@@ -0,0 +1,101 @@
...
+        if amplitude > 0.0:
+            rng = np.random.default_rng(output_seed)
+            noise = rng.normal(0, amplitude, len(audio.samples)).astype(np.float32)
+            output_samples: np.ndarray = np.clip(audio.samples + noise, -1.0, 1.0).astype(np.float32)
+        else:
+            output_samples = audio.samples
+
+        self._output_dir.mkdir(parents=True, exist_ok=True)
+        output_path = conventions.sample_file_path(self._output_dir, output_id, "wav")
+        await self._audio_writer.write(output_path, AudioData(samples=output_samples, sample_rate=audio.sample_rate))
```
jodavis, line 95: A new file shouldn't be written unless the sample is changed. Otherwise the manifest should just refer back to the original sample in the input directory. That should be how all stages are handled, so there isn't unnecessary copying of unmodified files. Essentially, if amplitude == 0.0, don't create anything new and just return `input`

ml/pipeline/speech/mic_noise_stage.py
jodavis: I meant only add the _mic### part if ### is not 0. But really, we shouldn't be creating a new file/ID at all if the content is unmodified.

ml/pipeline/speech/background_noise_stage.py
```diff
@@ -0,0 +1,135 @@
...
+
+        return AudioSample(
+            id=output_id,
+            seed=output_seed,
+            content_hash=content_hash,
+            path=Path(f"{output_id}.wav"),
+            parent_content_hash=parent_content_hash,
+            transcript=input_sample.transcript,
+            applied_values=applied_values,
+        )
```
jodavis, line 151: Similar to mic_noise_stage, if noise_volume == 0, don't create any of this. Just return `input` so that the manifest refers back to the original file. this avoids unnecessary copying of files.

jodavis (changes requested):

ml/pipeline/speech/background_noise_stage.py
jodavis-claude: Implemented. `NoiseProvider.list_files()` now returns `list[tuple[str, AudioData]]` instead of `list[Path]`. The filename string is kept alongside the `AudioData` for hash-stable variation selection in `_get_applied_values`. `BackgroundNoiseAugmentor._generate_output` looks up the pre-loaded noise `AudioData` by name from `list_files()` instead of reading the file from disk — no per-sample noise I/O. `BackgroundNoiseAugmentor` no longer requires `noise_dir` in its constructor (removed). `_DirectoryNoiseProvider` now accepts `sample_rate: int` and loads all WAV files eagerly via `librosa.load(sr=sample_rate)`, which resamples at load time so all noise audio matches the pipeline sample rate.

ml/pipeline/speech/background_noise_stage.py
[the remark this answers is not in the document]
jodavis-claude: Implemented with the `NoiseProvider` redesign (see thread at line 26). `_DirectoryNoiseProvider` now accepts `sample_rate: int` and resamples all noise files to that rate via `librosa.load(sr=sample_rate)` at construction time. `_generate_output` adds an explicit sample rate equality guard and raises `ValueError` with a diagnostic message if the noise provider returns audio at a different rate than the input — though with `_DirectoryNoiseProvider` this should never occur when configured correctly. `sample_rate` is now a pipeline-level parameter in `params.yaml` (defaulting to 16000) and `PipelineParams`, so all stages can share it.

ml/pipeline/speech/mic_noise_stage.py
jodavis-claude: Fixed. `_derive_id` now returns `input_sample.id` when `amplitude == 0.0` — no `_mic###` suffix. `_generate_output` returns `input_sample` unchanged when `amplitude == 0.0`, so no file is written and no I/O occurs. The unmodified sample is passed through the manifest as-is.

ml/pipeline/speech/mic_noise_stage.py
jodavis-claude: Fixed. When `amplitude == 0.0`, `_generate_output` returns `input_sample` directly — no audio is read, no file is written, and the manifest refers back to the original sample. Note: for this to work correctly with downstream stages, the downstream stage's `input_dir` must still be able to resolve `input_sample.path`. The DVC wiring (ADR-231) will need to account for this by pointing downstream stages at the appropriate directory when samples pass through unchanged.

ml/pipeline/speech/background_noise_stage.py
jodavis-claude: Fixed. When `noise_volume == 0.0`, `_generate_output` returns `input_sample` directly — no audio is read, no file is written, and the manifest refers back to the original sample. The same upstream-directory note applies: ADR-231 DVC wiring will need to ensure downstream stages can resolve `input_sample.path` when samples pass through unchanged.

jodavis (approved):

PR #235, opened 2026-06-23
ADR-229: Implement ModelTrainer, KerasBackend, and speech_10_train_model entry-point

jodavis-claude:
## Work item

ADR-229: Implement the `ModelTrainer` class, the `KerasBackend` protocol with its default TF/Keras implementation, and the `speech_10_train_model.py` entry-point — the stage that trains a CTC speech-to-text model from the featurised manifests produced by earlier pipeline stages.

## Changes

- `ml/pipeline/speech/model_trainer.py` — NEW. `KerasBackend` Protocol, `DefaultKerasBackend` (TF/Keras with deferred imports), and `ModelTrainer` class that filters manifests by `parent_id`, builds a `tf.data.Dataset`, and calls the backend to train and save.
- `ml/pipeline/stages/speech_10_train_model.py` — NEW. CLI entry-point following the `speech_07` pattern; accepts `--train-manifest-dir`, `--spectrogram-manifest-dir`, `--token-manifest-dir`, `--vocab-dir`, `--spectrogram-dir`, `--token-dir`, `--output-dir`.
- `ml/pipeline/stages/params.py` — MODIFIED. Added `TrainModelParams(epochs, batch_size)` dataclass and `train_model: TrainModelParams` field to `PipelineParams`.
- `ml/params.yaml` — MODIFIED. Added `train_model: epochs: 10, batch_size: 32` under `stages:`.
- `ml/test/pipeline/speech/test_model_trainer.py` — NEW. 15 pytest-style unit tests for `ModelTrainer` (filtering, backend call args, save/return path).
- `ml/test/pipeline/stages/test_params.py` — MODIFIED. Added `train_model` block to `_VALID_DATA` and 4 new `TestTrainModelParamsLoad` tests.
- `ml/test/pipeline/speech/test_token_stage.py` — MODIFIED. Fixed flaky mtime-based assertions in `TestSkipPath` — replaced with file content comparison to eliminate timer-resolution races on Windows NTFS.

## Design decisions

- **`KerasBackend` is a `Protocol`** — the default implementation defers `import tensorflow as tf` inside each method body (same deferred-import pattern as `librosa` in `SpectrogramStage`), keeping the module importable without TF installed.
- **Filtering is `ModelTrainer`'s responsibility** — the full combined spectrogram and token manifests are passed in; `train()` filters internally to entries whose `parent_id` is in the train split.
- **`ModelTrainer.train()` is synchronous** — the entry-point calls it directly without `asyncio.run()`.
- **`num_classes = vocab.ctc_blank_idx + 1`** — the `+1` is the CTC blank token, consistent with `VocabResult.ctc_blank_idx = len(phoneme_list)`.
- **`DefaultKerasBackend.build_ctc_model` uses `model.compile(loss='ctc')`** — Keras 3 / TF 2.x string alias for CTC loss; documented with a comment in the source.
- **Empty filtered set returns an empty list** from `_build_dataset` — avoids a TF import and works naturally with stub backend tests.

### DIFF
--- a/ml/pipeline/speech/_doc_speech.md
+++ b/ml/pipeline/speech/_doc_speech.md
@@ -1,6 +1,6 @@
 ## Summary: ml/pipeline/speech
 
-Speech pipeline stages for the OOP pipeline. The directory contains two cohorts:
+Speech pipeline stages for the OOP pipeline. The directory contains three cohorts:
 
 - **Augmentation stages** — `ModifierStage` subtypes that take an `AudioSample` (or
   `TextSample`) manifest and produce a derived `AudioSample` manifest with per-sample
@@ -12,6 +12,8 @@ Speech pipeline stages for the OOP pipeline. The directory contains two cohorts:
   different extension), and `_get_applied_values` returns `{}`.
 - **Splitter** — `SetManifestSplitter`, a plain (non-stage) class that partitions a
   fully-augmented `AudioSample` manifest into train/val/test subsets.
+- **Training** — `ModelTrainer`, a plain class that trains a CTC speech-to-text Keras
+  model from the featurised manifests and split produced by earlier pipeline stages.
 
 ## Stages
 
@@ -24,6 +26,7 @@ Speech pipeline stages for the OOP pipeline. The directory contains two cohorts:
 | [`spectrogram_stage.py`](spectrogram_stage.py) | `SpectrogramStage` | `AudioSample → SampleSpectrogram` (log-mel `.npy`) |
 | [`token_stage.py`](token_stage.py) | `TokenStage` | `AudioSample → SampleTokens` (phoneme token `.json`) |
 | [`set_splitter.py`](set_splitter.py) | `SetManifestSplitter` | Split augmented manifest into train/val/test |
+| [`model_trainer.py`](model_trainer.py) | `ModelTrainer` | Train CTC model from filtered spectrogram/token manifests |
 
 ## Key design decisions
 
@@ -105,3 +108,37 @@ sufficient and avoids adding a numpy dependency on a non-numerical operation.
 `train_manifest.json`, `val_manifest.json`, and `test_manifest.json` (the `_manifest`
 suffix is required by `conventions.split_manifest_path`). `AudioSample.id` values are
 preserved unchanged — no re-assignment in split outputs.
+
+### ModelTrainer
+
+**`KerasBackend` as the TF seam.** `ModelTrainer` accepts a `KerasBackend` Protocol
+and never imports TensorFlow directly. `DefaultKerasBackend` is the production
+implementation; it defers all `import tensorflow as tf` calls inside method bodies
+(`build_ctc_model` and `_build_dataset`) so the module is importable without TF
+installed (e.g. on CI without a GPU image). This mirrors the `TtsProvider`,
+`NoiseProvider`, and `AudioReader` patterns used elsewhere in the pipeline.
+
+**`ModelTrainer.train()` is synchronous.** The entry-point (`speech_10_train_model.py`)
+calls it directly — no `asyncio.run()` wrapper is needed.
+
+**Filtering is `ModelTrainer`'s responsibility.** `spectrogram_manifest` and
+`token_manifest` are the full combined manifests (all splits). `train_manifest` is
+the train-split subset produced by `SetManifestSplitter`. `ModelTrainer.train()`
+filters internally: only entries whose `parent_id ∈ {s.id for s in train_manifest.samples}`
+are included. Lookup dicts keyed by `parent_id` are built for both filtered sets before
+any file I/O begins.
+
+**Constructor-injected params.** `n_mels`, `time_steps`, `epochs`, and `batch_size` are
+passed to `ModelTrainer.__init__()`. The trainer does not read `PipelineParams` directly —
+the entry-point is responsible for bridging values from `PipelineParams.compute_spectrograms`
+and `PipelineParams.train_model` to the constructor.
+
+**Dataset construction is `_build_dataset`'s responsibility.** `ModelTrainer._build_dataset()`
+stacks spectrogram and token arrays into numpy arrays, constructs a `tf.data.Dataset`
+via `from_tensor_slices`, and applies `.batch(batch_size).prefetch(tf.data.AUTOTUNE)`.
+`KerasBackend.train()` receives an already-batched, already-prefetched dataset.
+
+**Empty-pair guard.** If filtering produces zero `(spectrogram, token)` pairs, a
+`WARNING` is logged before calling `build_ctc_model` / `train`. This surfaces
+misconfiguration (e.g. wrong manifest directories, split-manifest mismatch) that would
+otherwise cause Keras to silently fit on zero steps.
--- a/ml/pipeline/speech/_doc_speech.md
+++ b/ml/pipeline/speech/_doc_speech.md
@@ -27,6 +27,7 @@ Speech pipeline stages for the OOP pipeline. The directory contains three cohort
 | [`token_stage.py`](token_stage.py) | `TokenStage` | `AudioSample → SampleTokens` (phoneme token `.json`) |
 | [`set_splitter.py`](set_splitter.py) | `SetManifestSplitter` | Split augmented manifest into train/val/test |
 | [`model_trainer.py`](model_trainer.py) | `ModelTrainer` | Train CTC model from filtered spectrogram/token manifests |
+| [`manifest_filter.py`](manifest_filter.py) | `lookup_sample_triplets` | Match split, spectrogram, and token manifests into `(AudioSample, SampleSpectrogram, SampleTokens)` triples; shared by training, evaluation, and testing |
 
 ## Key design decisions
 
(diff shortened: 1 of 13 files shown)


ml/pipeline/speech/model_trainer.py
```diff
@@ -0,0 +1,275 @@
...
+            s.parent_id: s
+            for s in token_manifest.samples
+            if s.parent_id in train_ids
+        }
+
+        # Load arrays for samples that appear in both filtered manifests
+        pairs: list[tuple[np.ndarray, np.ndarray]] = []
+        for parent_id in spec_by_parent:
+            if parent_id not in tok_by_parent:
+                _logger.warning("No token sample for parent_id %s — skipping", parent_id)
```
jodavis-claude: **Priority 1 — Silent failure on zero training samples**

When filtering produces zero (spectrogram, token) pairs — either because `train_manifest` has no matching entries in both manifests, or because all pairs fail the token-lookup check — `_build_dataset` returns `[]` and `backend.train(model, [], epochs)` is called silently. Keras will run `model.fit([], epochs=N)` with zero steps and produce a trained model from nothing, with no indication that training was skipped.

Add a warning (or raise) here when `pairs` is empty, before calling `build_ctc_model` / `train`:

```python
if not pairs:
    _logger.warning(
        "ModelTrainer.train(): no (spectrogram, token) pairs found for train split — "
        "model will be trained on zero samples."
    )
```

Without this, a misconfigured pipeline (e.g. wrong `--train-manifest-dir` or a split-manifest/spectrogram-manifest mismatch) fails completely silently.

ml/pipeline/speech/model_trainer.py
```diff
@@ -0,0 +1,275 @@
...
+        """Fit the model and return per-epoch loss values."""
+        import tensorflow as tf  # deferred
+
+        history = model.fit(dataset, epochs=epochs)
+        loss_values: list[float] = history.history.get("loss", [])
+        return [float(v) for v in loss_values]
+
+    def predict(self, model: Any, dataset: Any) -> np.ndarray:
+        """Run model.predict() and return the raw output array."""
+        import tensorflow as tf  # deferred
```
jodavis-claude: **Priority 1 — Unused `import tensorflow as tf` in `predict` and `save`**

`predict` (line 145) and `save` (line 151) both import `tf` but never reference the `tf` name — they only call `model.predict(dataset)` and `model.save(str(path))` respectively. The `tf` variable is dead code in both methods.

The deferred-import pattern is justified only when `tf` is actually used in the method body (as in `_build_dataset` which calls `tf.data.Dataset.from_tensor_slices`). Remove the unused imports from `predict` and `save`. `train` also imports `tf` without using it — only `model.fit(...)` is called — remove that import too.

Note: calling `model.predict()` and `model.save()` on a TF model object does require TF to be loaded in the process, but that happens when `build_ctc_model` (which does use `tf`) is called earlier. The unused imports in these methods add noise without providing the intended guard.

ml/pipeline/speech/_doc_speech.md
```diff
@@ -0,0 +1,107 @@
...
+
+| Module | Class | Transform |
+|--------|-------|-----------|
+| [`tts_stage.py`](tts_stage.py) | `TtsSampleGenerator` | `TextSample → AudioSample` via edge_tts |
+| [`delay_stage.py`](delay_stage.py) | `DelayAugmentor` | `AudioSample → AudioSample` (silence padding) |
+| [`background_noise_stage.py`](background_noise_stage.py) | `BackgroundNoiseAugmentor` | `AudioSample → AudioSample` (environmental noise mix) |
+| [`mic_noise_stage.py`](mic_noise_stage.py) | `MicrophoneNoiseAugmentor` | `AudioSample → AudioSample` (Gaussian mic noise) |
+| [`spectrogram_stage.py`](spectrogram_stage.py) | `SpectrogramStage` | `AudioSample → SampleSpectrogram` (log-mel `.npy`) |
+| [`token_stage.py`](token_stage.py) | `TokenStage` | `AudioSample → SampleTokens` (phoneme token `.json`) |
+| [`set_splitter.py`](set_splitter.py) | `SetManifestSplitter` | Split augmented manifest into train/val/test |
```
jodavis-claude, line 28: **Priority 4 — `_doc_speech.md` does not cover `ModelTrainer`**

`model_trainer.py` is added to `ml/pipeline/speech/` by this PR but the document does not mention it. The intro describes only two cohorts (Augmentation stages, Featurisation stages); the Stages table stops at `set_splitter.py`; there is no section for `ModelTrainer`.

The following should be added:

1. **Stages table**: add a row for `model_trainer.py` / `ModelTrainer` — e.g. "Train CTC model from filtered spectrogram/token manifests".

2. **New section — `ModelTrainer`**: document the key design decisions:
   - `KerasBackend` as the TF seam (Protocol + `DefaultKerasBackend` with deferred imports), mirroring the `TtsProvider` / `NoiseProvider` / `AudioReader` patterns.
   - `ModelTrainer.train()` is synchronous — the entry-point calls it directly without `asyncio.run`.
   - Filtering responsibility: `spectrogram_manifest` and `token_manifest` are full combined manifests; `ModelTrainer` filters internally by `parent_id ∈ {s.id for s in train_manifest.samples}`.
   - `n_mels`, `time_steps`, `epochs`, `batch_size` are constructor parameters (not read from params internally) — the entry-point bridges them from `PipelineParams`.
   - Dataset construction (`batching`, `prefetching`) is `ModelTrainer._build_dataset`'s responsibility; `KerasBackend.train` receives an already-prepared dataset.

Per `CONTRIBUTING.md`: "When new subsystems are added, they should include a documentation file" and "When designs are updated, documents should be updated to match."

jodavis-claude (commented):
## ADR-229 Review — ModelTrainer, KerasBackend, speech_10_train_model

**Exit criteria check:**
- `KerasBackend` protocol + `DefaultKerasBackend` with all five methods (`build_ctc_model`, `train`, `predict`, `save`, `load`): implemented correctly.
- `ModelTrainer.train()`: filters by `parent_id`, constructs `tf.data.Dataset` with batching/prefetching, calls `KerasBackend.train` then `save`, returns model path: implemented correctly.
- `speech_10_train_model.py` entry-point with all required CLI args: implemented correctly.
- `TrainModelParams` dataclass + `stages.train_model` in `params.py` and `params.yaml`: implemented correctly.
- Unit tests (15 tests in `test_model_trainer.py` + 4 in `test_params.py`): all required test classes present and correctly structured.
- Build/tests pass: confirmed by context file validate log.

**Issues found:**

**Priority 1 — Two correctness issues:**

1. **Silent zero-sample training** (`model_trainer.py` line 228): When the filter produces zero (spectrogram, token) pairs, `_build_dataset` returns `[]` and `backend.train(model, [], epochs)` is called with no warning. Keras will silently fit on zero steps. A `_logger.warning(...)` should be emitted before `build_ctc_model` / `train` are called in the empty case.

2. **Unused `import tensorflow as tf`** in `DefaultKerasBackend.train` (line 137), `predict` (line 145), and `save` (line 151): `tf` is imported but never referenced in any of these three method bodies. The deferred-import convention is meaningful only when `tf.*` is actually used. Remove these three dead imports.

**Priority 4 — Documentation gap:**

3. **`_doc_speech.md` omits `ModelTrainer`** (line 26): `model_trainer.py` is a new module in `ml/pipeline/speech/` but has no entry in the Stages table and no design section. The intro does not mention training as a third cohort. The key decisions (KerasBackend as TF seam, synchronous training, filtering responsibility, constructor-injected params) need to be documented. Per `CONTRIBUTING.md`: new subsystems require a documentation file, and designs must be updated to match.

**Style notes (not blocking):**
- `DefaultKerasBackend` does not explicitly declare `(KerasBackend)` as a base — structural typing makes this valid, but explicit inheritance is used by `EdgeTtsProvider` and similar classes elsewhere for clarity.
- `dvc.yaml` only defines stages through `speech_01` — downstream stages including `speech_10` are not yet wired. Not blocking since other stages are in-flight, but note that the entry-point cannot be exercised via `dvc repro` yet.

ml/pipeline/speech/model_trainer.py
jodavis-claude: Applied. Added `_logger.warning(...)` immediately before `build_ctc_model` / `train` when `pairs` is empty. Added `TestZeroSampleWarning::test_warning_logged_when_no_pairs` to verify the warning is emitted when `train_manifest` has no matching entries in the spectrogram/token manifests. Committed in `ADR-229: Warn when ModelTrainer.train() produces zero (spec, token) pairs; remove dead TF imports`.

ml/pipeline/speech/model_trainer.py
jodavis-claude: Applied. Removed the unused `import tensorflow as tf` from `DefaultKerasBackend.train()`, `predict()`, and `save()`. The `load()` method correctly retains its deferred import since it calls `tf.keras.models.load_model`. Committed in `ADR-229: Warn when ModelTrainer.train() produces zero (spec, token) pairs; remove dead TF imports`.

ml/pipeline/speech/_doc_speech.md
jodavis-claude: Applied. Updated `_doc_speech.md` with:
1. Added `ModelTrainer` row to the Stages table.
2. Updated the intro to list a fourth cohort ("Training").
3. Added a `### ModelTrainer` design section covering: `KerasBackend` as the TF seam (Protocol + deferred imports), synchronous `train()`, filter-by-`parent_id` responsibility, constructor-injected params, dataset construction ownership, and the empty-pair guard.

Committed in `ADR-229: Update _doc_speech.md to cover ModelTrainer`.

ml/pipeline/speech/_doc_speech.md
```diff
@@ -105,3 +108,37 @@ sufficient and avoids adding a numpy dependency on a non-numerical operation.
...
+`token_manifest` are the full combined manifests (all splits). `train_manifest` is
+the train-split subset produced by `SetManifestSplitter`. `ModelTrainer.train()`
+filters internally: only entries whose `parent_id ∈ {s.id for s in train_manifest.samples}`
+are included. Lookup dicts keyed by `parent_id` are built for both filtered sets before
+any file I/O begins.
+
+**Constructor-injected params.** `n_mels`, `time_steps`, `epochs`, and `batch_size` are
+passed to `ModelTrainer.__init__()`. The trainer does not read `PipelineParams` directly —
+the entry-point is responsible for bridging values from `PipelineParams.compute_spectrograms`
+and `PipelineParams.train_model` to the constructor.
```
jodavis: This is too much implementation detail for the documentation files. If anything, it should be mentioned more generally, as in "entry-point scripts are responsible for reading params and providing parameters to stage implementations" as a general rule, not specific to this type.

ml/pipeline/speech/_doc_speech.md
```diff
@@ -105,3 +108,37 @@ sufficient and avoids adding a numpy dependency on a non-numerical operation.
...
+and never imports TensorFlow directly. `DefaultKerasBackend` is the production
+implementation; it defers all `import tensorflow as tf` calls inside method bodies
+(`build_ctc_model` and `_build_dataset`) so the module is importable without TF
+installed (e.g. on CI without a GPU image). This mirrors the `TtsProvider`,
+`NoiseProvider`, and `AudioReader` patterns used elsewhere in the pipeline.
+
+**`ModelTrainer.train()` is synchronous.** The entry-point (`speech_10_train_model.py`)
+calls it directly — no `asyncio.run()` wrapper is needed.
+
+**Filtering is `ModelTrainer`'s responsibility.** `spectrogram_manifest` and
```
jodavis: We'll need similar functionality for evaluation and testing, so there should be a common utility that takes a filtered manifest, the spectrogram manifest, and the token manifest, and returns a data structure that is a List(AudioSample, SampleSpectrogram, SampleTokens) for processing. That utility should be thoroughly unit tested and shared with training, evalution, and testing.

ml/pipeline/speech/model_trainer.py
```diff
@@ -0,0 +1,275 @@
...
+# ---------------------------------------------------------------------------
+# DefaultKerasBackend — production TF/Keras implementation
+# ---------------------------------------------------------------------------
+
+
+class DefaultKerasBackend:
+    """Production KerasBackend that wraps TensorFlow/Keras.
+
+    All ``import tensorflow as tf`` calls are deferred to individual method
+    bodies so that importing this module does not require TF to be installed.
```
jodavis: Why not move DefaultKerasBackend to a separate module, and import `tf` in that module? Then it won't have to be deferred because that whole module would not be loaded by unit tests.

ml/pipeline/speech/model_trainer.py
```diff
@@ -0,0 +1,275 @@
...
+
+_logger = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# KerasBackend protocol
+# ---------------------------------------------------------------------------
+
+
+class KerasBackend(Protocol):
```
jodavis: I think this protocol should be moved to a separate module from `model_trainer`. The same abstraction will be needed for evaluation and testing, so it should be in a sharable location.

ml/pipeline/speech/model_trainer.py
```diff
@@ -0,0 +1,275 @@
...
+from pipeline.core.manifest import Manifest
+from pipeline.core.sample import AudioSample, SampleSpectrogram, SampleTokens
+from pipeline.intent.vocab_computer import VocabResult
+from pipeline.stages import conventions
+
+_logger = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# KerasBackend protocol
```
jodavis: Because this is a generic protocol, it should have a name that doesn't mention the "Keras" implementation. Maybe something like "MachineLearningModelBuilder" that is then implemented by "TensorflowModelBuilder" instead of "DefaultWhatever" which isn't descriptive of what the implementation is doing.

ml/pipeline/speech/model_trainer.py
```diff
@@ -0,0 +1,275 @@
...
+
+
+class KerasBackend(Protocol):
+    """Injectable abstraction over TensorFlow/Keras.
+
+    The default implementation (DefaultKerasBackend) defers all TF imports
+    inside method bodies. Test stubs can implement this protocol without TF.
+    """
+
+    def build_ctc_model(self, num_classes: int, n_mels: int, time_steps: int) -> Any:
```
jodavis: Make this a more strongly-typed API by returning a type like "MachineLearningModel(Protocol)" instead of Any. Move the `train`, `predict`, and `save` functions to `MachineLearningModel`.

jodavis (changes requested):

ml/pipeline/speech/_doc_speech.md
jodavis-claude: Applied. Removed the specific constructor-parameter paragraph and replaced it with a general note: "Entry-point scripts bridge PipelineParams to stage implementations — this pattern applies uniformly across all pipeline stages." The specific param names belong in the source, not the doc.

ml/pipeline/speech/_doc_speech.md
jodavis-claude: Applied. Created `manifest_filter.py` with `lookup_sample_triplets(split_manifest, spectrogram_manifest, token_manifest)` returning `list[tuple[AudioSample, SampleSpectrogram, SampleTokens]]`. The utility is split-agnostic — callers pass whichever split they need. Added 14 unit tests in `test_manifest_filter.py` covering: all matched, empty split, extra entries excluded, missing spectrogram (skipped + warning), missing token (skipped + warning), missing both, and partial matches. `ModelTrainer` now delegates to this utility instead of filtering inline. Documented in `_doc_speech.md`.

ml/pipeline/speech/model_trainer.py
jodavis-claude: Applied. Moved `DefaultKerasBackend` (renamed `TensorflowModelBuilder`) to `tensorflow_backend.py` with a module-level `import tensorflow as tf`. Unit tests import only `model_trainer.py` and `ml_model.py`, neither of which touches TF — so TF is never loaded during testing.

ml/pipeline/speech/model_trainer.py
jodavis-claude: Applied. Moved `KerasBackend` (renamed `MachineLearningModelBuilder`) and the new `MachineLearningModel` protocol to `ml_model.py`. Both are importable by training, evaluation, and testing stages without pulling in TensorFlow or any stage-specific code.

ml/pipeline/speech/model_trainer.py
jodavis-claude: Applied. Renamed `KerasBackend` → `MachineLearningModelBuilder` and `DefaultKerasBackend` → `TensorflowModelBuilder`. The protocol name describes the role (builds models); the implementation name describes what it uses (TensorFlow).

ml/pipeline/speech/model_trainer.py
jodavis-claude: Applied. Added `MachineLearningModel` Protocol to `ml_model.py` with `train(dataset, epochs) -> list[float]`, `predict(dataset) -> np.ndarray`, and `save(path) -> None`. `MachineLearningModelBuilder.build_ctc_model` now returns `MachineLearningModel` instead of `Any`. `ModelTrainer` calls `model.train(dataset, epochs)` and `model.save(path)` directly on the returned model. `TensorflowModel` in `tensorflow_backend.py` wraps a `tf.keras.Model` and implements this protocol.

jodavis (approved):

PR #248, opened 2026-07-28
ADR-339: ML pipeline core data model (Sample hierarchy, Manifest, ManifestStore)

jodavis-claude:
## Work item

`ADR-339` — build the unified `Sample`/`Manifest[S]`/`ManifestStore` data model in `ml/pipeline/core/`: the typed sample hierarchy, generic manifest collection, and JSON (schema v1) round-trip store that every later ML pipeline stage (ADR-340 onward) will serialise through.

## Changes

- `ml/pipeline/core/__init__.py` — new, empty package marker for the `core/` subpackage.
- `ml/pipeline/core/sample.py` — new. `Sample` base dataclass (`name`, `content_hash`, `applied_values`, `parent_name`, `parent_content_hash`, all defaulted); `SampleWithPath(Sample)` adds `path`; `TextSample(Sample)` adds `content`/`label`, with `__post_init__` setting `name = content_hash`; `AudioSample(SampleWithPath)` adds `transcript`/`label`/`seed`; `SampleSpectrogram`/`SampleTokens` (both `SampleWithPath`, path-based, no `seed`).
- `ml/pipeline/core/manifest.py` — new. `Manifest[S]` (classic `Generic[S]`/`TypeVar` style; `by_name`/`by_content_hash` dict-backed lookups; `add()` raises `ValueError` on duplicate names; `__len__`/`__iter__` for enumeration). `ManifestStore` (`write`/`read`; JSON schema v1 — `schema_version`, `sample_type`, `samples`; `write` raises `ValueError` on empty manifest or mixed sample types; `read` type-dispatches off the persisted `sample_type` field via a `_SAMPLE_TYPES_BY_NAME` registry).
- `ml/test/pipeline/__init__.py`, `ml/test/pipeline/core/__init__.py` — new, empty test package markers mirroring the source layout.
- `ml/test/pipeline/core/test_sample.py`, `test_manifest.py`, `test_manifest_store.py` — new, test-first unit tests covering all of the above.

## Design decisions

- `Manifest.__len__`/`__iter__` were added beyond the Component Breakdown's literal text — needed by `ManifestStore.write` to detect an empty manifest and inspect sample types, judged as within `Manifest[S]`'s own collection responsibility.
- Fixed a gap found during self-review: the TDD trio's first `ManifestStore.read` implementation wrote a `sample_type` field but still required the caller to pass the type explicitly rather than dispatching off it. Replaced with a `_SAMPLE_TYPES_BY_NAME` registry so deserialization dispatches purely from the persisted JSON (separate commit).
- `SampleSpectrogram`/`SampleTokens` are path-based (extend `SampleWithPath`, no extra fields), consistent with `AudioSample` and the `ModifierStage` output-directory pattern — resolves a known ambiguity per the task brief's own recommendation.
- All six dataclasses are plain (non-frozen), every field defaulted, per the task brief's recommendation (no spec requirement for immutability).
- Open ambiguity, not resolved in this PR: no Python test runner (`pytest`) is wired into `scripts/validate.sh` yet, so this PR's unit tests cannot be verified by the project's standard build/test quality gate script. Recommend a small follow-up task adding `pytest` to `ml/requirements.txt` and a `scripts/validate-ml-tests.sh`/`.cmd`.
- No E2E/Gherkin scenarios — this repo's only E2E harness targets the C# WPF/Blazor UI; not applicable to this Python ML pipeline code.

## Testing completed

- 15 unit tests, all passing (`python3 -m pytest test/ -v` from `ml/`).
- `mypy --strict pipeline test` passes with zero issues.

### DIFF
--- a/ml/pipeline/core/manifest.py
+++ b/ml/pipeline/core/manifest.py
@@ -1,0 +1,42 @@
+"""Generic typed collection of `Sample` objects, keyed by `name` and `content_hash`."""
+
+from __future__ import annotations
+
+from typing import Generic, Iterator, TypeVar
+
+from pipeline.core.sample import Sample
+
+S = TypeVar("S", bound=Sample)
+
+
+class Manifest(Generic[S]):
+    """A named/content-hash-indexed collection of one concrete `Sample` subtype.
+
+    Enforces name uniqueness within the manifest — two samples sharing a
+    `name` is a construction-time error, since `name` is what backs on-disk
+    filenames and cross-manifest joins (`parent_name`).
+    """
+
+    def __init__(self, samples: list[S] | None = None) -> None:
+        self._by_name: dict[str, S] = {}
+        self._by_content_hash: dict[str, S] = {}
+        for sample in samples or []:
+            self.add(sample)
+
+    def add(self, sample: S) -> None:
+        if sample.name in self._by_name:
+            raise ValueError(f"Duplicate sample name: {sample.name!r}")
+        self._by_name[sample.name] = sample
+        self._by_content_hash[sample.content_hash] = sample
+
+    def by_name(self, name: str) -> S:
+        return self._by_name[name]
+
+    def by_content_hash(self, content_hash: str) -> S:
+        return self._by_content_hash[content_hash]
+
+    def __len__(self) -> int:
+        return len(self._by_name)
+
+    def __iter__(self) -> Iterator[S]:
+        return iter(self._by_name.values())
--- a/ml/pipeline/core/manifest.py
+++ b/ml/pipeline/core/manifest.py
@@ -33,6 +33,8 @@ class Manifest(Generic[S]):
     def add(self, sample: S) -> None:
         if sample.name in self._by_name:
             raise ValueError(f"Duplicate sample name: {sample.name!r}")
+        if sample.content_hash in self._by_content_hash:
+            raise ValueError(f"Duplicate sample content_hash: {sample.content_hash!r}")
         self._by_name[sample.name] = sample
         self._by_content_hash[sample.content_hash] = sample
 
--- a/ml/pipeline/core/manifest.py
+++ b/ml/pipeline/core/manifest.py
@@ -7,9 +7,13 @@ import json
 from pathlib import Path
 from typing import Any, Generic, Iterator, TypeVar
 
-from pipeline.core.sample import Sample
+from pipeline.core.sample import AudioSample, Sample, SampleSpectrogram, SampleTokens, TextSample
 
 S = TypeVar("S", bound=Sample)
+
+_SAMPLE_TYPES_BY_NAME: dict[str, type[Sample]] = {
+    cls.__name__: cls for cls in (TextSample, AudioSample, SampleSpectrogram, SampleTokens)
+}
 
 
 class Manifest(Generic[S]):
@@ -65,6 +69,7 @@ class ManifestStore:
         }
         path.write_text(json.dumps(payload))
 
-    def read(self, path: Path, sample_type: type[S]) -> Manifest[S]:
+    def read(self, path: Path) -> Manifest[Sample]:
         payload = json.loads(path.read_text())
-        return Manifest[S]([sample_type(**data) for data in payload["samples"]])
+        sample_type = _SAMPLE_TYPES_BY_NAME[payload["sample_type"]]
+        return Manifest[Sample]([sample_type(**data) for data in payload["samples"]])
--- a/ml/pipeline/core/manifest.py
+++ b/ml/pipeline/core/manifest.py
@@ -2,7 +2,10 @@
 
 from __future__ import annotations
 
-from typing import Generic, Iterator, TypeVar
+import dataclasses
+import json
+from pathlib import Path
+from typing import Any, Generic, Iterator, TypeVar
 
 from pipeline.core.sample import Sample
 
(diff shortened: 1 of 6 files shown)


ml/pipeline/core/manifest.py
```diff
@@ -0,0 +1,75 @@
...
+        self._by_name: dict[str, S] = {}
+        self._by_content_hash: dict[str, S] = {}
+        for sample in samples or []:
+            self.add(sample)
+
+    def add(self, sample: S) -> None:
+        if sample.name in self._by_name:
+            raise ValueError(f"Duplicate sample name: {sample.name!r}")
+        self._by_name[sample.name] = sample
+        self._by_content_hash[sample.content_hash] = sample
```
jodavis-claude, line 39: **Correctness / fault tolerance:** `add()` enforces name uniqueness (raises `ValueError`) but silently overwrites `_by_content_hash[sample.content_hash]` if two samples with *different* names share the same `content_hash` — the earlier sample becomes permanently unreachable via `by_content_hash()` with no error or warning. Given the spec's framing of `content_hash` as the field that "actually determines sample identity" for downstream skip/regenerate decisions (`_spec_OopPipeline.md`, "Content hash determines sample identity"), a silent, undetected collision here seems like exactly the kind of failure mode this review is supposed to catch (no silent failures). Since this class is the shared data model every later stage builds on, consider either raising `ValueError` on a duplicate `content_hash` too (mirroring the existing duplicate-name check), or — if collisions are intentionally tolerated — documenting that `by_content_hash` resolves to the last-added sample and why that's safe.

ml/pipeline/core/manifest.py
```diff
@@ -0,0 +1,75 @@
...
+        payload: dict[str, Any] = {
+            "schema_version": 1,
+            "sample_type": sample_type_name,
+            "samples": [dataclasses.asdict(sample) for sample in samples],
+        }
+        path.write_text(json.dumps(payload))
+
+    def read(self, path: Path) -> Manifest[Sample]:
+        payload = json.loads(path.read_text())
+        sample_type = _SAMPLE_TYPES_BY_NAME[payload["sample_type"]]
```
jodavis-claude: **Fault tolerance / consistency:** an unrecognized `sample_type` in the JSON (corrupted file, hand-edited manifest, or a future schema version this registry doesn't know about yet) raises a bare `KeyError: '<name>'` here, whereas every other failure mode in `ManifestStore` (empty manifest, mixed types) raises a descriptive `ValueError`. Consider wrapping this lookup to raise `ValueError(f"Unknown sample type in manifest: {payload['sample_type']!r}")` for a clearer, consistent failure message. Related: `schema_version` is written but never checked on read — not blocking for a v1-only pipeline, but worth keeping in mind once a v2 format exists.

jodavis-claude (commented):
Reviewed against ADR-339's exit criteria (Sample hierarchy, Manifest[S], ManifestStore) and `ml/_spec_OopPipeline.md`'s Key Design Decisions / Component Breakdown. All four exit-criteria bullets are met, `mypy --strict pipeline test` and `python3 -m pytest test/ -v` both pass locally (15/15). Tests follow AAA structure with clear Arrange/Act/Assert comments and one behavior per test.

Two correctness/fault-tolerance issues in `manifest.py` worth addressing before merge (see inline comments): a silent duplicate-`content_hash` overwrite in `Manifest.add`, and a raw `KeyError` (instead of a descriptive `ValueError`, consistent with the rest of `ManifestStore`) when `read()` encounters an unrecognized `sample_type`.

Note (non-blocking, already flagged by the developer in the PR description): `pytest` still isn't wired into `scripts/validate.sh`, so this PR's tests can't be verified by the project's standard quality-gate script yet. Worth a quick follow-up task, but out of scope for ADR-339's own exit criteria.

ml/pipeline/core/manifest.py
jodavis-claude: Agreed — fixed. Manifest.add() now raises ValueError on a duplicate content_hash too, mirroring the existing duplicate-name check, rather than silently overwriting the earlier sample in _by_content_hash. Added a failing-first regression test using AudioSample (whose name is independent of content_hash, unlike TextSample) to exercise the collision path that was not caught by the existing duplicate-name test. See commit 163a0e3.

ml/pipeline/core/manifest.py
jodavis-claude: Agreed — fixed. ManifestStore.read() now wraps the registry lookup and raises ValueError(f"Unknown sample type in manifest: {sample_type_name!r}") instead of a bare KeyError, consistent with the other ManifestStore failure modes. Added a failing-first test that writes a manifest JSON with an unrecognized sample_type and asserts the ValueError. Left the schema_version-not-validated-on-read note as-is per your own call that it is not urgent for a v1-only pipeline. See commit 6e5611b.

ml/pipeline/core/manifest.py
jodavis-claude: Confirmed — verified the new `content_hash` collision check in `Manifest.add()` raises `ValueError` before either dict is mutated, and the new AudioSample-based regression test exercises it correctly. All 17 tests pass, mypy --strict clean. Resolving.

ml/pipeline/core/manifest.py
jodavis-claude: Confirmed — `ManifestStore.read()` now raises a descriptive `ValueError` on an unrecognized `sample_type` instead of a bare `KeyError`, consistent with the other failure modes, and the new regression test covers it. Resolving.

jodavis-claude (commented):
## Sign-off review

Both prior review threads verified as correctly addressed:

1. **`Manifest.add()` silent content_hash overwrite** — fixed with a `ValueError` check on duplicate `content_hash` (mirroring the existing duplicate-name check), placed before either backing dict is mutated. Regression test `test_add_duplicate_content_hash_different_names_raises_value_error` uses `AudioSample` to correctly isolate the content_hash collision from the name-uniqueness path. Verified.
2. **`ManifestStore.read()` bare `KeyError` on unknown sample_type** — fixed with a `try`/`except KeyError` that re-raises a descriptive `ValueError`, consistent with the class's other failure modes. Regression test `test_read_unrecognized_sample_type_raises_value_error` covers it. Verified.

Ran the full suite locally: 17/17 pytest tests pass, `mypy --strict pipeline test` clean.

No new issues found in the files touched by this fix pass (`ml/pipeline/core/manifest.py`, `ml/test/pipeline/core/test_manifest.py`, `ml/test/pipeline/core/test_manifest_store.py`).

One pre-existing note carried forward for awareness, not blocking this PR: Known Ambiguity #1 (no `pytest` wiring into `scripts/validate.sh`/`ml/requirements.txt`) remains unresolved, as flagged in the original task brief and implementation summary — Python unit tests still cannot be verified through the project's standard quality-gate script. Worth a small follow-up task before the ML test suite grows further.

Approving for hand-off.

jodavis-claude:
Abandoning this task's PR — resetting ADR-191's tasks to re-run the pipeline with recent workflow fixes. No action needed; a fresh task will supersede this.

